PR #5, opened 2022-08-30
Slight cleanups based on another pair of eyes

windiana42:
# Checklist

### DIFF
--- a/src/pydiverse/pipedag/core/task.py
+++ b/src/pydiverse/pipedag/core/task.py
@@ -157,15 +157,15 @@ class Task:
         args = self._bound_args.args
         kwargs = self._bound_args.kwargs
 
-        def task_result_visitor(x):
+        def task_result_mapper(x):
             if isinstance(x, Task):
                 return x.value
             elif isinstance(x, TaskGetItem):
                 return x.value
             return x
 
-        args = deep_map(args, task_result_visitor)
-        kwargs = deep_map(kwargs, task_result_visitor)
+        args = deep_map(args, task_result_mapper)
+        kwargs = deep_map(kwargs, task_result_mapper)
 
         with TaskContext(task=self):
             result = self.fn(*args, **kwargs)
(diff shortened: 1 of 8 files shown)


src/pydiverse/pipedag/core/task.py
```diff
@@ -157,15 +157,15 @@ def _run(self):
         args = self._bound_args.args
         kwargs = self._bound_args.kwargs
 
-        def task_result_mutator(x):
+        def task_result_visitor(x):
```
NMAC427, line 160: Is this really a visitor? Maybe `retrieve_task_result` would be a better name because the function doesn't just visit the args / kwargs, it modifies them.

src/pydiverse/pipedag/materialize/store.py
```diff
@@ -117,7 +117,7 @@ def dematerialize_task_inputs(
 
         ctx = RunContext.get()
 
-        def dematerialize_mutator(x):
+        def dematerialize_visitor(x):
```
NMAC427, line 120: Not really a visitor, because it maps its inputs; But can't come up with a better name either.

NMAC427 (commented):
`deepmutate` should also be renamed to `deep_map` in pydiverse transform.

src/pydiverse/pipedag/materialize/store.py
windiana42: something that maps its inputs is a mapper

windiana42: I suggest to distinguish: visitors (output is ignored, no side effects), mappers (output is the main result, no side effects), mutators (side effects on the objects visited)
There may be a grey area between "output is ignored" and "output is main result".

src/pydiverse/pipedag/core/task.py
windiana42: also a mapper

NMAC427 (approved):

PR #6, opened 2022-08-31
Test and support SQL Server as TableStore

windiana42:
SQL Server has quite some specific aspects to consider when supporting it as a SQL backend. It is not just a different sqlalchemy connection string. Getting sqlalchemy and SQL Server working together is a particular challenge since SQL Server supports queries across *Databases* which is handled by sqlalchemy as the schema parameter with a pattern like "database.schema". It is even trickier to ensure correct database/schema quoting because '[database.schema]' is not correct. Furthermore, it does not work to select the database in the connection and to use the schema parameter just for the SQL Server schema because sqlalchemy will occationally run 'USE schema' which won't work.

# Checklist


windiana42:
@NMAC427 for better debugging stack traces, I think it is safe to assume that flow.run() is called from the main thread/process and thus raising exceptions should be at least possible as a development option:
```python
...
            res = engine.run(flow=self, **kwargs)
        if not res.successful and isinstance(res.underlying, Exception):
            # make sure proper stack trace is printed
            raise res.underlying
        return res
```

src/pydiverse/pipedag/backend/table/util/sql_ddl.py
```diff
@@ -59,6 +59,12 @@ def __init__(self, name, schema, if_exists=False):
 @compiles(CreateSchema)
 def visit_create_schema(create: CreateSchema, compiler, **kw):
     schema = compiler.preparer.format_schema(create.name)
+    if compiler.dialect.name == "mssql":
```
NMAC427, line 62: The correct way to specify dialect specific compilation rules is to use `@compiles(CreateSchema, "mssql")` ([reference](https://docs.sqlalchemy.org/en/14/core/compiler.html#dialect-specific-compilation-rules)).

windiana42:
Another change, I plan to do is to open connections of an engine as a function of the TableStore. This allows us to run database specific calls at the beginning of connects/transactions like `SET TRANSACTION ISOLATION LEVEL READ UNCOMMITTED`.

NMAC427:
> @NMAC427 for better debugging stack traces, I think it is safe to assume that flow.run() is called from the main thread/process and thus raising exceptions should be at least possible as a development option:

This probably shouldn't be the default behavior, because most engines (for example Prefect) do error handling themself. I probably would implement it as an argument `flow.run(raises=True)`.

src/pydiverse/pipedag/backend/table/util/sql_ddl.py
windiana42: Will do. Will it handle both annotations correctly such that we have a default implementation and some target specific ones?

NMAC427: Yes. If it can't find a specific one, it will fall back to the generic default implementation.

windiana42:
> This probably shouldn't be the default behavior

At the beginning all code breaks. So I like to get the best debugging support I can before I actually can think of reasonable pipeline orchestration or later monitoring. So my preference is that fail-fast is the default. The question is whether we have adoption disadvantages from this decision because people don't realize they can change the option when going to DAG monitoring.
Another alternative would be to make sure the full stacktrace is displayed in all cases (which it wasn't before bis change).

NMAC427:
Fail fast only works with the sequential engine. Prefect just logs the exception and continues executing the DAG.

I feel like logging the exception gives exactly the same level of debugger support as raising it. The upside of not raising an exception (and instead just logging it) would be that the result object (and thus all run metadata) could still be inspected. This would, for example, allow you to visualize the DAG and show exactly which task failed and which were successful by assigning colors to them.

windiana42:
> I feel like logging the exception gives exactly the same level of debugger support as raising it.

That is at least not the case with the current implementation. I just received one stack frame and not all stack frames of the exception.
I also would like unit tests to fail on error when just calling f.run(). But I see the trade-off between testing concerns and orchestration/monitoring concern. Let's keep the decision of the default mode for later when we can poll a larger user base. I am happy to let fail_fast default to false for now.
I guess another discussion is whether fail_fast should be a parameter to run() or a field in the config object. Let's also keep this discussion for later when we have more clarity about piperun/pipetest.

windiana42:
@EnchoMishinevQC I started out trying to support two modes with SQL Server 1) working with databases as schemas and 2) working with schemas as schemas. It seems like getting the schemas as schemas mode off the ground has trouble with schema renaming: https://www.sqlservercentral.com/articles/renaming-a-schema-in-sql-server

It can be done but probably not simply by a sqlalchemy compiler function. But maybe we will support other means of transactionality like readonly-views. Then it is probably also feasible to implement the moving of all objects from one schema to the next.

windiana42:
Tests don't work yet with sqlserver, but we are getting closer...

windiana42:
@NMAC427 somehow the display of errors is strange. I like to quickly skim my stacktrace from bottom to top to find the place where my code calls library code. And the output is filled with very long printouts of code between stack frames. I prefer debuggers to automatically injected printf debugging.

windiana42:
`raise e` prints much nicer stack traces (for me) than `raise StageError from e`. I suspect that the problem can be fixed by configuration but I don't know what needs to be configured.
```python
        try:
            yield True
        except Exception as e:
            # Failed initializing stage
            self._request("exit_init_stage", stage.id, False)
            #raise StageError from e
            raise e
```

windiana42:
@NMAC427 in my opinion we are ready to merge SQL Server capabilities to main branch

windiana42:
We still need to test multiple table store backends with github actions, but I would see this as a separate PR.

src/pydiverse/pipedag/backend/table/util/sql_ddl.py
```diff
@@ -78,46 +137,160 @@ def visit_drop_schema(drop: DropSchema, compiler, **kw):
...
+        "EXCEPT",
+        "INTERSECT",
+    ]:
+        regex = re.compile(marker, re.IGNORECASE)
+        for match in regex.finditer(select_sql):
+            match_start = match.span()[0]
+            prev = select_sql[0:match_start]
+            # ignore marker in subqueries in select columns
+            if prev.count("(") == prev.count(")"):
+                into_point = match_start
```
EnchoMishinevQC, line 240: Perhaps this should pick the smallest `match_start` among all markers, as otherwise it's not really getting the first marker. E.g. `SELECT * FROM t ORDER BY val` would plug the `INTO` before the `ORDER BY` because it's later in the marker list, resulting in an invalid query.

EnchoMishinevQC (approved):
I think this is in a good enough state to merge and test further. Can probably be refactored in some nicer way once it's more stable.

src/pydiverse/pipedag/backend/table/util/sql_ddl.py
windiana42: @EnchoMishinevQC this is indeed a bug. I added a failing test and will fix it:
```
E             - Select 1 INTO a.b.c FROM A GROUP BY x
E             ?                     -------
E             + Select 1 FROM A INTO a.b.c GROUP BY x
E             ?         +++++++
```

PR #10, opened 2022-09-20
Implement raw SQL statements as @materialize task output for migrating legacy code to pipedag 

windiana42:
This feature focuses on mssql dialect since it is common to have length scripts with dynamic SQL running on SQL-Server which are not easy to separate out in single SELECT statements. That said, it should always be the goal to separate out SQL code in separate SELECT statements once they are running in pydiverse.pipedag. This feature should make pydiverse.pipedag adoption simpler and more continuous than a big bang migration. 

# Checklist


windiana42:
We couldn't get the open source package `pytsql` running for our T-SQL scripts (drop function caused trouble to the grammer), thus we resorted to the following workaround. We don't need to make it too nice to work with lengthy dynamic SQL scripts:
```python
            # workaround: pytsql failed the scripts in this repo (DROP FUNCTION)
            last_use = None
            for stmt in re.split(r"\bGO\b", sql, flags=re.IGNORECASE):
                if stmt.strip() == "":
                    # allow GO at end of script
                    continue
                if re.match(r"\bUSE\b", stmt.strip()):
                    last_use = stmt
                with self.engine.connect() as conn:
                    if last_use is not None:
                        self.execute(last_use, conn=conn)
                    self.execute(stmt, conn=conn)
```

windiana42:
@EnchoMishinevQC may I ask you for a code review on this PR as well?
The user point of view of this change can already be seen in: https://github.com/Quantco/pipedag_demo_tsql

NMAC427:
Can you implement some tests for this?

windiana42:
> Can you implement some tests for this?

Sorry, I should have. I tested with a non-open-source Repo.
I will add extensive RawSQL tests in #19 with and without pytsql.

src/pydiverse/pipedag/backend/table/base.py
```diff
@@ -226,6 +248,45 @@ def store_table_lazy(self, table: Table):
             )
         )
 
+    def store_raw_sql(self, raw_sql: RawSql):
+        """Lazily sores a table in the associated commit stage
```
EnchoMishinevQC, line 252:
```suggestion
        """Lazily stores a table in the associated commit stage
```

src/pydiverse/pipedag/backend/table/sql.py
```diff
@@ -464,6 +673,15 @@ def materialize(cls, store, table: Table[pd.DataFrame], stage_name):
...
         )
 
+    @classmethod
+    def execute_raw_sql(
+        cls,
+        store,
+        sql: str,
+        stage_name: str,
+    ):
+        SQLAlchemyTableHook.execute_raw_sql(store, sql, stage_name)
```
EnchoMishinevQC, line 683: I'm a bit confused as to why we have to implement this in the `PandasTableHook`. What's the situation in which this makes sense?

EnchoMishinevQC (approved):
Looks good to me overall. Will be very helpful for new adopters. I've left just a couple of questions for my clarification.

I'll separately look into the issues preventing `pytsql` from being used here.

src/pydiverse/pipedag/backend/table/base.py
windiana42: will fix

src/pydiverse/pipedag/backend/table/util/sql_ddl.py
```diff
@@ -127,6 +154,7 @@ def visit_create_schema(create: CreateSchema, compiler, **kw):
 
 @compiles(DropSchema)
 def visit_drop_schema(drop: DropSchema, compiler, **kw):
+    _ = kw
```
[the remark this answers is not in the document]
windiana42: They are are way to tell the IDE to ignore unused symbol warnings. Essentially not using '_' is the only symbol that is tolerated by default.

src/pydiverse/pipedag/backend/table/sql.py
windiana42: I guess this code is to blame: base.py: 197:
```python
         # take any registered table hook since we only care about the materialization target
        hook = self._REGISTERED_TABLES[0]
        hook.execute_raw_sql(self, raw_sql.sql, raw_sql.stage.transaction_name)
```

windiana42: The whole concept of tables hooks does not make that much sense for RawSQL execution. It is meant for storing tables individually: base.py: 187:
```python
        hook = self.get_m_table_hook(type(table.obj))
```
We could either implement a complete different set of hooks for raw sql execution, or we implement the same logic for all hooks of a table store. Currently the latter choice was taken.

Issue #12, opened 2022-09-22
Cache invalidation of source nodes with help of user


windiana42:
We still need a way to allow the user help in cache invalidating source nodes. There was the idea to pass a custom cache validity checking function to @materialize. The problem of this approach is that there might be still user specific input needed which is available during wiring of tasks in the flow. So I am currently favoring the approach to add special @materialize tasks at the beginning which always run (i.e. lazy=true will do) and emit a CacheInfo(xxx) object which is used to compute the cache validity hash. It could look like this:
```python
@materialize(input_type=sa.Table, lazy=True)
def source_valid(landing_zone_schema: str):
    config = ConfigContext.get()
    store = config.store.table_store
    meta = sa.MetaData()
    meta.reflect(bind=store.engine, schema=landing_zone_schema)
    tbl = sa.Table("main_table", meta)
    last_timestamp = sa.select([sa.func.max(tbl.c.timestamp)]).fetchall()[0][0] 
    return CashInfo(last_timestamp)
```
The CacheInfo object can be passed on to other @materialize tasks creating a dependency chain by that.

It is even possible to pass multiple arguments to CacheInfo which all will be traversed and hashed with `deep_visit()`. And it is possible to pass SQLAlchemy tables to the output which effectively copies the tables into the stage where source_tables() task is part of:
```python
@materialize(input_type=sa.Table, lazy=True)
def source_tables(landing_zone_schema: str):
    config = ConfigContext.get()
    store = config.store.table_store
    meta = sa.MetaData()
    meta.reflect(bind=store.engine, schema=landing_zone_schema)
    last_timestamps = [sa.select([sa.func.max(tbl.c.timestamp)]).fetchall()[0][0] for tbl in meta.tables]
    return CashInfo(landing_zone_schema, last_timestamps, [tbl.name for tbl in meta.tables]), [Table(tbl) for meta.tables]
```

windiana42:
@NMAC427 @EnchoMishinevQC  what do you think?

NMAC427:
For comparison, this is the version I proposed:

```python
## Table Example
def get_meta(schema: str):
    config = ConfigContext.get()
    store = config.store.table_store
    meta = sa.MetaData()
    meta.reflect(bind=store.engine, schema=schema)
    return meta

def schema_cache(schema: str):
    meta = get_meta()
[...]

PR #13, opened 2022-10-06
Added source cache invalidation (#12)

andysham:
Draft solution to #12, will open to merge once some unit tests are added.

### DIFF
--- a/src/pydiverse/pipedag/materialize/core.py
+++ b/src/pydiverse/pipedag/materialize/core.py
@@ -151,7 +151,7 @@ class MaterializationWrapper:
         # Compute the cache key for the task inputs
         input_json = store.json_encode(bound.arguments)
         if task.cache is not None:
-            input_json += json.dumps(task.cache(*args, **kwargs))
+            input_json += store.json_encode(task.cache(*args, **kwargs))
         cache_key = store.compute_task_cache_key(task, input_json)
         task.cache_key = cache_key
 
--- a/src/pydiverse/pipedag/materialize/core.py
+++ b/src/pydiverse/pipedag/materialize/core.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import functools
 import inspect
+import json
 from functools import partial
 from typing import Callable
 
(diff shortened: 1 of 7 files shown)


windiana42 (approved):
Simple integration. Thank you! I will add a test and then we are good.

windiana42:
@andysham FYI: I added a testcase (I also needed to implement flow.debug_collect() to check the output of a flow.run().

NMAC427:
I just took the `flow.debug_collect()` logic, cleaned it up a bit and moved it into the `Result` class.

On another note: Why do you want to merge this change into prefect_1 branch instead of main?

NMAC427:
One thing I still think is worth considering a bit more is how exactly the `cache` argument should function given some possible use cases:

- A task that should always run.
    - Very useful for the first task in a flow. E.g. a task that just returns a list of files in a directory.
    - Possible Syntax:
        - `@materialize(cache=None)`
        - `@materialize(cache=Never)` where `Never` is a function provided by pipedag that just returns a UUID.
- A task whose inputs might change, but where it can easily be determined that the task will produce the same output.
    - This would require there to be a mechanism to ignore the task inputs and only use the cache function.

src/pydiverse/pipedag/materialize/core.py
```diff
@@ -144,6 +150,8 @@ def __call__(self, *args, **kwargs):
 
         # Compute the cache key for the task inputs
         input_json = store.json_encode(bound.arguments)
+        if task.cache is not None:
+            input_json += json.dumps(task.cache(*args, **kwargs))
```
NMAC427, line 154: This should use `store.json_encode` instead of `json.dumps` to ensure that the json produced is canonical (-> all dictionary keys are sorted).

windiana42:
> On another note: Why do you want to merge this change into prefect_1 branch instead of main?

I find it easier to order PRs and to merge them in this order. Reduces merge conflicts in case of more significant refactorings. But I am happy to rebase this PR on main, push first, and then rebase/force push the other open PRs. Ideally the problem vanishes by faster code review / main merges.

windiana42:
> * `@materialize(cache=None)`
> 
> * `@materialize(cache=Never)` where `Never` is a function provided by pipedag that just returns a UUID.

I prefer the cache=never syntax (lowercase since it is a function and not a Class). I currently focus on providing value add for simple pipedag adoption (multiple instances / pipetest). So I might leave low hanging fruit like this for later or s.o. else to pick it up...

I actually think that ignoring fresh input at the source might also be a useful option for running a DAG instance. I am currently seeing three parameters there: instance-ID, whether to run it per-user or team-shared, whether to ignore fresh input on the source.

src/pydiverse/pipedag/materialize/core.py
windiana42: done

Issue #17, opened 2022-11-27
support creation of multiple DAG instances with one configuration file (prerequisite for pipetest)


windiana42:
The power of pipedag is that it should make it easy to create multiple instances of a DAG working with small/medium/large data inputs and per-user as well as team shared materialization. However, we still need to define and implement a configuration format that allows specifying multiple DAG instances. I would like to keep the binding to pipetest loose by just defining tags that may or may not be picked up by pipetest to schedule DAG instance executions with a unit testing framework.
An tricky point is that sometimes one likes to develop on some stages of a pipeline which are technically cache-invalid in case of new source input. But for development purposes one might want to ignore this fresh input.

windiana42:
@NMAC427 @verticube here is my first attempt to convert the existing pipedag configuration (https://github.com/pydiverse/pydiverse.pipedag/blob/main/tests/pipedag.toml, https://github.com/pydiverse/pydiverse.pipedag/blob/main/docs/reference/config.rst) to one that may be used for defining the execution of multiple instances:
```
name: pipedag_tests
database_connections:
  postgres:
    # Postgres: this can be used after running `docker-compose up`
    engine: "postgresql://sa:Pydiverse23@127.0.0.1:6543/{database}"

  mssql:
    # SQL Server: this can be used after running `docker-compose up`
    engine: "mssql+pyodbc://sa:PidyQuant27@127.0.0.1:1433/master?driver=ODBC+Driver+18+for+SQL+Server&encrypt=no"
    # schema_prefix: "master."  # SQL Server needs database.schema (uncomment only on of prefix and suffix)
    schema_prefix: "{database}_"  # SQL Server needs database.schema (uncomment only on of prefix and suffix)
    schema_suffix: ".dbo"   # Alternatively SQL Server databases can be used as schemas with .dbo default schema

instances:
  default:
    # listen-interface for pipedag context server which synchronizes some task state during DAG execution
    network_interface: "127.0.0.1"
    # classes to be materialized to table store even without pipedag Table wrapper (we have loose coupling between
    # pipedag and pydiverse.transform, so consider adding 'pydiverse.transform.Table' in your config)
    auto_table: ["pandas.DataFrame", ]
    # abort as fast a possible on task failure and print most readable stack trace
    fail_fast: true

    table_store:
      class: "pydiverse.pipedag.backend.table.SQLTableStore"

      # Postgres: this can be used after running `docker-compose up`
      database_connection: postgres
[...]

PR #18, opened 2022-11-27
Configure multiple instances of a pipe-DAG (issue #17)

windiana42:
# Checklist


windiana42:
@NMAC427 this is a considerable change to configuration concepts. So I would like to get some feedback from you.

src/pydiverse/pipedag/util/deep_map.py
```diff
@@ -30,7 +31,7 @@ def deep_map(x, callable: Callable, memo=None):
     else:
         y = callable(x)
 
-    # If is its own copy, don't memoize.
+    # If is its own copy, don't memorize.
```
NMAC427, line 34: `memoize` is the correct word:
https://github.com/python/cpython/blob/main/Lib/copy.py#L174
https://en.wikipedia.org/wiki/Memoization

src/pydiverse/pipedag/context/context.py
```diff
@@ -97,41 +94,16 @@ class ConfigContext(BaseAttrsContext):
...
-        return PipeDAGStore(
-            table=table_store,
-            blob=blob_store,
-        )
-
-    def get_lock_manager(self) -> BaseLockManager:
-        return load_instance(self.config_dict["lock_manager"])
+    store: PipeDAGStore
+    lock_manager: BaseLockManager
+    engine: Engine
```
NMAC427, line 99: This won't work when using a multiprocessing / multinode based engine (e.g. prefect with dask), because you can't pickle those objects.
That's basically the reason why I came up with a config file in the first place. I needed a mechanism to instantiate a database connection, etc. in different processes.

tests/pipedag.yaml
```diff
@@ -0,0 +1,99 @@
+name: pipedag_tests
```
NMAC427, line 1: Must be moved into the instance section, else the ConfigContext has `name` set to `None`.

NMAC427:
I'm still not convinced that associating a specific "Instance Config" with a flow is in any way necessary or even desirable. It should be possible to run the same flow with different configurations without having to reinstantiate the flow multiple times. A flow should just be a graph consisting of stages and task - nothing else.

I much prefer having a mechanism for specifying which "instance" / "config variation" to use as part of `flow.run`. I believe this results in more elegant code both on the user side, as well as on the side of pipedag / pipetest. Implementing this should just require changing `ConfigContext.from_file` so that it also accepts an instance name in addition to a file path (in addition to the new yaml based config reader).

By checking if flow.run is being called inside a ConfigContext, and using it instead of creating a now one, it should also be relatively easy to test a flow on different instances:

```python
@pipetest.run(instances=["full-prefect", "full-dagster"])
def test_foo():
    flow = get_flow()
    flow.run()
```

The `pipetest.run` decorator just has to open the correct config context manager for each instance:

```python
# This pseudo code is far from correct, but demonstrates how the same test
# function could be run with different instances without passing around
# config objects as described in
# https://github.com/pydiverse/pydiverse.pipedag/issues/17#issuecomment-1328210380
def pipetest_run(instances):
    for instance in instances:
        with ConfigContext.from_file(instance=instance):
            test_function()
```

src/pydiverse/pipedag/util/deep_map.py
QuantCo-mr-T: Oops

src/pydiverse/pipedag/context/context.py
QuantCo-mr-T: Then these objects should parse and verify the input first, but open/close connections on start/stop.

QuantCo-mr-T:
I would also use the instance config for some flow modifying attributes. Of course the creator of the flow could extract it from the config and feed it as separate inputs to the flow creating function. This is actually a pretty standard case where filtering mini-DBs is most easy from an already loaded full-DB.

windiana42:
@NMAC427  I tried to encorporate most of the feedback. I don't like flow.run() receiving instance parameters even though this would still work as an additional change. I prefer to have a configuration for one pipedag-Instance from which one can use flow_attributes during flow creation which can also be passed into flow.run(). With statement wrappers are nice, but they hide quasi-global communication. So I won't use them much outside of flow.run(). But I have no problem for other people to use it, so I also implemented that mode.

windiana42:
We might still have a naming confusion. `InstanceConfig` is the configuration for all instances at once. `ConfigContext` is the configuration for one pipe-DAG-Instance. Any naming suggestions here?

NMAC427:
Looking through your code, I don't see why the class `InstanceConfig` is even needed.
The only two things it does is provide a way to initialize a ConfigContext and it is passed to the `_init_conf_` functions where `instance_config.get_flow_name()` is accessed.

The latter use, however, is incorrect. This is how the `name` config parameter (which doesn't refer to the flow name - the flow name is set when calling Flow.__init__) is defined in the documentation:

> The name of the project. **Optional**
>
> Currently, the main purpose of the project name is as a namespace for the lock manager.
> That way you can use the same manager (e.g. Zookeeper) for two projects that run on different databases without causing any issues.

This means that if we run a flow on two different "instances" (btw. I'm still not entirely sure what you are referring to by instance) they should have different "names". This means that the `name` config parameter should be moved to the instance section instead.
Consequently, it doesn't make sense to pass an `InstanceConfig` to `_init_conf_`, because we need an explicit config (given by `ConfigContext`) to do proper namespacing.

NMAC427:
You can achieve the exact same effect by using ConfigContext without the need for an instance config object:

```python
# Version 1 - Dynamic flow creation without context-local state
config = ConfigContext(file=None, instance="default")

with Flow("FLOW") as flow:
    with Stage("simple_flow_stage1"):
        if not config.attrs["copy_filtered_input"]:
            ...

flow.run(cfg)


# Version 2 - Dynamic flow creation with context-local state
def init_flow(cfg):
    config = ConfigContext.get()

    with Flow("FLOW") as flow:
        with Stage("simple_flow_stage1"):
            if not config.attrs["copy_filtered_input"]:
                ...

    return flow

with ConfigContext(file=None, instance="default"):
    flow = init_flow()
    flow.run()


# Version 3 - Static flow creation with context-local state
@materialize
def input():
    config = ConfigContext.get()
    if not config.attrs["copy_filtered_input"]:
        ...

with Flow("Flow") as flow:
    x = input()

config = ConfigContext(file=None, instance="default")
flow.run(config)
# or
flow.run(instance="default")
```

Additionally I would rename `flow_attributes` to `attrs` because a config file doesn't necessarily refer to a specific flow (e.g. all the flows in the pipedag test cases use the same config file).

QuantCo-mr-T:
I like to decouple reading yaml configuration + potentially overriding some aspects of that via command line args / environment variables from selecting an instance out of the config. So InstanceConfig is this intermediate representation of the yaml that may be used in various ways.

QuantCo-mr-T:
I called flow_attributes like that because the are interpreted by the flow. attrs seems too unspecific to me to see that they are something radically different than the other sections of an instance. I am still open to renaming though.

windiana42:
> This means that if we run a flow on two different "instances" (btw. I'm still not entirely sure what you are referring to by instance) they should have different "names". This means that the `name` config parameter should be moved to the instance section instead.

I would like to keep the name on instance level. I see the highest value in this name to be a flow identifier for human consumption that might also be used when looking at DAG orchestration UI. There should be one name for multiple instances in the hierarchy level below. For ZooKeeper, I suggest to create extra name components to disambiguate different run instances. Locking is most important for a specific database connection + schema name. I am unsure whether we should prevent a collision of mssql://.../master, schema=mydb.stage1 and mssql://.../mydb, schema=stage1

windiana42:
@NMAC427 I think we have three aspects that have influence on the config/flow interaction which are overlapping in meaning:
1) I want to make it easy to think about a DAG as one thing under which I can have DAG instances (different input data, different materialization schemas, per user/team shared, ...).
2) Since we don't have a hard coupling of a flow and a pipedag configuration, multiple flows may be executed with the same config
3) The user may want to run only some stages of a DAG instance/flow combination and he might even develop exploratively in the sense of a large variety of flows under one flow name

You seem to prefer putting the flow top level, mixing it mainly with technical configuration that is potentially shared across flows. I like to put the DAG top level and see different flows executed in a DAG context just like executing substages of one flow. Essentially the DAG may be a forest of many flows / stage-selections executed with the same pipedag config. If flows overlap in a schema name, it is suggested that everything up to and including this schema name is identical code. If not, those flows will compete for cache resources and might evict each other. Branch-based CI executions with a branch independent database is such a case of resource contention. This can either be solved by the cache managing more than two schemas, or by branch specific databases with a database purging logic.

docs/reference/config.rst
```diff
@@ -1,25 +1,178 @@
...
+
+The `pipedag.yaml` file is used to configure pipedag.
 
 When executing a flow, pipedag searches for the config file in the following directories:
 
 - The current working directory
 - Any parent directories of the working directory
 - The user folder
 
+**Attention: pipedag config files may be used for code execution under different user ID.**
```
cmdevries, line 16:
```suggestion
**Attention: pipedag config files may be used for code execution under a different user ID.**
```

docs/reference/config.rst
```diff
@@ -1,25 +1,178 @@
...
+Pipedag may be instructed to read yaml files with key-value pairs and sending content to arbitrary
```
cmdevries, line 17:
```suggestion
Pipedag may be instructed to read yaml files with key-value pairs and send content to arbitrary
```

src/pydiverse/pipedag/backend/blob.py
```diff
@@ -90,14 +91,25 @@ class FileBlobStore(BaseBlobStore):
...
+        blob_store_connection: str | None = None,
+    ):
         self.base_path = os.path.abspath(base_path)
-        config = ConfigContext.get()
-        if config.name is not None:
-            project_name = normalize_name(config.name)
-            self.base_path = os.path.join(self.base_path, project_name)
+        self.blob_store_connection = blob_store_connection  # for debug output
+        self.instance_id = cfg.instance_id
+        os.makedirs(os.path.join(self.base_path, self.instance_id), exist_ok=True)
```
cmdevries, line 107: Have you considered using Pathlib? https://docs.python.org/3/library/pathlib.html

src/pydiverse/pipedag/backend/lock.py
```diff
@@ -240,32 +249,46 @@ class ZooKeeperLockManager(BaseLockManager):
...
+            self.client.start()
+            atexit.register(lambda: self.dispose())
+        self.client.add_listener(self._lock_listener)
 
-    def close(self):
-        self.client.stop()
-        self.client.close()
+        self.logger.debug("opened lock manager", instance_id=self.instance_id)
+
+    def dispose(self):
```
cmdevries, line 274: I am not familiar with the API design choices here but was there a reason `acquire` and `dispose` were chosen rather than say using `__enter__` and `__exit__` for Python with statements?

cmdevries (commented):
Interesting to see pipedag and pipetest continue development 👍

src/pydiverse/pipedag/backend/lock.py
windiana42: These locks are managed by ServerRunContext(). No user should ever call acquire/dispose here. We can't use `with` statements for this kind of locking since we might dynamically move the locks as we move through the flow.

src/pydiverse/pipedag/backend/blob.py
windiana42: I exempt sphinx `conf.py` from this rule. I prefer to leave copy-paste solution as is...

windiana42: done

src/pydiverse/pipedag/core/result.py
```diff
@@ -26,15 +29,24 @@ def get(self, task: Task | TaskGetItem, as_type: type = None) -> Materializable:
             If no type is specified, the input type of the task is used.
         :return: The results of the task.
         """
+        if not self.successful:
+            logger = structlog.getLogger()
+            logger.warning(
+                "Attention: getting tables from unsuccessful run is unreliable!"
```
NMAC427, line 35: It is unreliable in general. If you run the testcases with the flag `-n auto` (run tests in parallel), many tests that call `result.get()` fail, because a different test has already locked and cleared the stage.

tests/pipedag.yaml
windiana42: we chose different solution

src/pydiverse/pipedag/materialize/util/json.py
```diff
@@ -25,26 +27,32 @@ def json_default(o):
...
+            "name": o.name,
+            "cache_keys": o.cache_keys,
         }
     if isinstance(o, Stage):
         return {
             PIPEDAG_TYPE: PIPEDAG_TYPE_STAGE,
             "name": o.name,
         }
-    if isinstance(o, Blob):
+    if isinstance(o, PipedagConfig):
```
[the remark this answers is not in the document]
windiana42: You are right. So far I am only using PipedagConfig in the flow to transport data from one flow to another. This is a rather static thing which typically is not changed by environment variables or url_attrs_file. Of course the database connection could be changed once in a while.

But does this really matter if PipedagConfig is never emitted from a task (which actually is a really awkward case)? I would expect that the source node's cache function would get it always fresh and thus can decide whether the new database connection yields the same hash.

We could simply fail if PipedagConfig is returned from a task.

src/pydiverse/pipedag/core/result.py
windiana42: I was thinking about offering to keep stage locks beyond flow.run(). I would also need a similar feature for pipetest: check whether stage output is cache valid and if so, give me a handle to locks that I may release after test is finished.
But this is probably an advanced feature which I wouldn't implement right away.

cmdevries (approved):

PR #20, opened 2022-12-13
Integrate pytsql optionally for mssql dialect

windiana42:
pytsql offers some nice functionality like forwarding PRINT statements. But it does not support all kinds of tsql input. So we make it optional.

# Checklist


windiana42:
Will merge this PR to master without Review, since we are currently heavily refactoring caching logic in main branch with the risk of lots of merge conflicts for this PR. The tests are failing on this PR after merging in main, because there were tests added to main branch showing failing behavior of previous changes to cache logic. This PR was green before.

Issue #22, opened 2022-12-15
Fix cache invalidation for Flow.run(ignore_fresh_input=True)


windiana42:
We decided to swtich from the cache_keys concept currently implemented to storing input_hash+cache_fn_hash for tasks and to rely on the fact that the transport from output_json to next task's input_hash is always going through the database. And cache hashes in the database only change if the task producing a certain input was invalidated.

windiana42:
@NMAC427 which hashes should we use for memoization? If we just use the input_hash, rerunning the same task will yield the exact same result for the same input. Like this it is for example impossible to write a random number generator as task and to expect that it will yield different outputs within one stage. On the other hand if it has a `@materialize(cache=fn)` function, then it appears like the multiple instances of the task would be called in the exact same moment producing the exact same result. I think this behavior is ok.

```python
        ctx = RunContext.get()
        with ctx.task_memo(task, task.input_hash) as (success, memo):
```

windiana42:
I think this is the place where task cache key and query str are combined into one hash: base.py:394
```python
        return compute_cache_key("LAZY-TABLE", table.cache_key, query_str)
```
This essentially means that when we store a cache_key in a Table object, then it must be the correct combination of input_hash, version, and cache_fn_hash according to ignore_fresh_input parameter. The name Table.cache_key, however, may be misleading for this kind of use.

windiana42:
The `Flow.run(ignore_fresh_input)` option is read in two places.
1) for Eager:
```python
    def retrieve_task_metadata(self, task: MaterializingTask) -> TaskMetadata:
        # noinspection PyUnresolvedReferences
        ignore_fresh_input = RunContext.get().ignore_fresh_input
        try:
            with self.engine.connect() as conn:
                result = (
                    conn.execute(
                        self.tasks_table.select()
                        .where(self.tasks_table.c.stage == task.stage.name)
                        .where(self.tasks_table.c.version == task.version)
                        .where(self.tasks_table.c.input_hash == task.input_hash)
                        .where(self.tasks_table.c.cache_fn_hash == task.cache_fn_hash if not ignore_fresh_input else sa.literal(True))
                        .where(self.tasks_table.c.in_transaction_schema.in_([False]))
                    )
                    .mappings()
                    .one_or_none()
                )
```
2) For Lazy:
```python
            # Do the materialization
            if isinstance(x, (Table, RawSql, Blob)):
                x.cache_key = task.get_effective_cache_key(ctx)

[...]

PR #24, opened 2022-12-17
Fix cache behaviour (see #22, #23)

NMAC427:
This fixes the incorrect behaviour of `flow.run(ignore_fresh_input=True)`.


windiana42 (approved):
We already discussed this intensively. It definitely improves the broken cache invalidation logic in main branch. We may still do some more tweaks to find a good abstraction for integrating a Caching Algorithm that may evolve and have different behaviors depending on options.

PR #28, opened 2022-12-29
Speed up and expand testing

NMAC427:
This PR add two features:
- It adds a instance fixture that allows running tests on many different instances.
- It adds a pytest plugin that allows executing independent tests in parallel (on my computer this results in a 3.5x speedup when running with `--mssql`).

This PR results in a few test cases related to ibm_db2 failing. This is because there is something wrong with how we handle ibm_db2. Sadly I can't debug this because the IBM Db2 odbc drivers don't support Apple Silicone Macs.


NMAC427:
As a future step, we should also include instances that test prefect, and the different stage commit techniques.

PR #29, opened 2022-12-29
Sql refactoring

NMAC427:
This PR cleans up how we handle different sql dialects. It introduces a `engine_dispatch` decorator that dispatches to the correct function implementation based on the current engine's dialect. This allows for significantly less nested and more readable code.

I also changed how `DropSchema` with `cascade=True` gets handled for IBM Db2. Instead of manually having to drop everything in the schema before calling `DropSchema`, the `DropSchema` visitor now inspects the schema and creates an sql query that first drops all tables and only then drops the schema.

Finally, I also replaced the `mcr.microsoft.com/mssql/server:2019-latest` docker image with `mcr.microsoft.com/azure-sql-edge`, because the latter also works on Apple Silicon Macs.

### DIFF
--- a/src/pydiverse/pipedag/backend/table/util/sql_ddl.py
+++ b/src/pydiverse/pipedag/backend/table/util/sql_ddl.py
@@ -362,7 +362,6 @@ def visit_drop_database(drop: DropDatabase, compiler, **kw):
     raise NotImplementedError(
         f"Disable for now for safety reasons (not yet needed): {ret}"
     )
-    # return ret
 
 
 def _visit_create_obj_as_select(create, compiler, _type, kw, *, prefix="", suffix=""):
--- a/src/pydiverse/pipedag/backend/table/util/sql_ddl.py
+++ b/src/pydiverse/pipedag/backend/table/util/sql_ddl.py
@@ -45,10 +45,15 @@ class CreateSchema(DDLElement):
 
 
 class DropSchema(DDLElement):
-    def __init__(self, schema: Schema, if_exists=False, cascade=False):
+    def __init__(self, schema: Schema, if_exists=False, cascade=False, *, engine=None):
+        """
+        :param engine: Used if cascade=True but the database doesn't support cascade.
+        """
         self.schema = schema
         self.if_exists = if_exists
         self.cascade = cascade
+
+        self.engine = engine
 
 
 class RenameSchema(DDLElement):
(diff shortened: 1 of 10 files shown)


src/pydiverse/pipedag/backend/table/util/sql_ddl.py
```diff
@@ -45,11 +45,16 @@ def __init__(self, schema: Schema, if_not_exists=False):
 
 
 class DropSchema(DDLElement):
-    def __init__(self, schema: Schema, if_exists=False, cascade=False):
+    def __init__(self, schema: Schema, if_exists=False, cascade=False, *, engine=None):
```
windiana42, line 48: @NMAC427 I find it a bit strange to introduce engine calls in the sqlalchemy compiler framework. I rather had in mind that we develop an abstraction around conn.execute() calls which works similar to the compiler framework in the sense that code for different dialects lives separately but it can issue multiple conn.execute() calls in the course of one "operation" (i.e. dropping a schema or swapping a schema). But for the moment, this change is definitely a step in the right direction.

src/pydiverse/pipedag/backend/table/util/sql_ddl.py
```diff
@@ -252,19 +256,52 @@ def visit_drop_schema(drop: DropSchema, compiler, **kw):
...
+    text = ["DROP SCHEMA"]
+    if drop.if_exists:
+        text.append("IF EXISTS")
+    text.append(schema)
+    statements.append(" ".join(text))
+
+    return ";\n".join(statements)
+
+
+@compiles(DropSchema, "ibm_db_sa")
```
windiana42, line 293: This looks like `@compiles(DropSchema, "ibm_db_sa")` is declared twice, now. If this is true, a comment explaining what this means might help.

src/pydiverse/pipedag/backend/table/util/sql_ddl.py
```diff
@@ -428,17 +461,15 @@ def insert_into_in_query(select_sql, database, schema, table):
     )
 
 
-# noinspection SqlDialectInspection
 @compiles(CopyTable)
 def visit_copy_table(copy_table: CopyTable, compiler, **kw):
     from_name = compiler.preparer.quote_identifier(copy_table.from_name)
     from_schema = compiler.preparer.format_schema(copy_table.from_schema.get())
-    query = sa.text(f"SELECT * FROM {from_schema}.{from_name}")
+    query = sa.select("*").select_from(sa.text(f"{from_schema}.{from_name}"))
```
windiana42, line 468: Is it new, that `sa.select("*")` works? I would have expected the need for `sa.select([sa.text("*")])`. But of course the former version looks much nicer.

src/pydiverse/pipedag/backend/table/util/sql_ddl.py
windiana42: Can't passing `engine` to `DropSchema` be avoided by using the `@engine_dispatch` concept?

windiana42 (approved):
looks good to me

PR #32, opened 2023-01-01
Extract cache management functionality out of core.py

windiana42:
This should pave the way to allow implementing n-slot caching without too much change to core.py.

# Checklist


windiana42:
@NMAC427 this is just a first try to separate out the caching functionality in a separate class with an info object moved around by the calling code. There is certainly fine-tuning necessary which functionality goes where but it should give an idea about changes I intend.

windiana42:
I will merge to main on test green in order to release 0.2.0 as the first version usable in real projects.
@NMAC427 sorry, that you didn't find the time to "intervene" here. I am sure we will change this a lot more. But I think it is a step in the right direction.

PR #33, opened 2023-01-07
Create Primary Keys and Indexes

windiana42:
We already had a primary_key parameter to Table constructor, but it was not implemented. And it is non-trivial to get that to work with postgres, mssql, and ibm_db_sa...

# Checklist


windiana42:
I will merge to main on test green in order to release 0.2.0 as the first version usable in real projects.

PR #37, opened 2023-01-22
add some tests for tasks using pydiverse.transform

windiana42:
Also added regex way to display table store URL with password obfuscation for easier debugging.

# Checklist


src/pydiverse/pipedag/backend/table/sql.py
```diff
@@ -61,11 +61,19 @@ class SQLTableStore(BaseTableStore):
...
+        groups = re.match(re_pw, engine_url)
+        engine_url_no_pw = "<unknown URL pattern>"
+        if len(groups.groups()) == 4 and len(groups.groups()[1]) > 0:
+            engine_url_no_pw = groups[1] + "***" + groups[3] + groups[4]
+        return cls(engine_url, engine_url_no_pw, **config)
 
     def __init__(
         self,
         engine_url: str,
+        engine_url_no_pw: str,
```
NMAC427, line 76: This shouldn't be an argument. It should get constructed inside of __init__ based on the engine_url.

src/pydiverse/pipedag/backend/table/sql.py
```diff
@@ -110,6 +120,7 @@ def __init__(
         self.metadata_schema = self.get_schema(self.METADATA_SCHEMA)
 
         self._init_database(engine_url, create_database_if_not_exists)
+        self.engine_url_no_pw = engine_url_no_pw
```
NMAC427, line 123: `engine_url_no_pw` only gets used once for a log message. Why do we save it as an attribute?

pyproject.toml
```diff
@@ -55,6 +56,7 @@ zookeeper = ["kazoo"]
 docs = ["Sphinx", "sphinx-rtd-theme", "sphinxcontrib-apidoc"]
 mssql = ["pyodbc", "pytsql"]
 ibm_db2 = ["ibm-db", "ibm-db-sa"]
+pdtransform = ["pydiverse-transform"]
```
NMAC427, line 59: I don't think this should be an extra as it isn't a dependency of pipedag, but a dependency of usercode intead.
For development purposes, adding transform to the `[tool.poetry.group.dev.dependencies]` makes more sense.

pytest.ini
```diff
@@ -6,6 +6,7 @@ markers =
     postgres: a test that requires postgres
     mssql: a test that requires mssql
     ibm_db2: a test that requires ibm_db2
+    pdtransform: a test that requires pydiverse-transform
```
NMAC427, line 9: I don't think this is necessary, as the pydiverse transform tests don't take a lot of time. They're only supposed to test that the `PydiverseTransformTableHook` is properly implemented.

To automatically skip tests that can't succeed because a module isn't installed, I propose to use [pytest.importskip](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#skipping-on-a-missing-import-dependency) instead.

NMAC427 (dismissed):
As this mostly test that `PydiverseTransformTableHook` is implemented correctly, I propose that we create explicit tests for all TableHooks instead. Otherwise, there is nothing special about how pipedag treats transform.

In case of `PydiverseTransformTableHook`, for example, this would mean that we test that we can materialize and dematerialize as `PandasTableImpl` and `SQLTableImpl`, that table names get passed along correctly, and that the lazy query string gets generated correctly.

src/pydiverse/pipedag/backend/table/sql.py
```diff
@@ -61,11 +61,19 @@ class SQLTableStore(BaseTableStore):
...
         engine_url = config.pop("url")
-        return cls(engine_url, **config)
+        # TODO :we might need user support or SQL-URL library support to do this
+        #  reliably:
+        re_pw = r"([^:]*://[/]*[^:/@]*:)([^@/]+)(@|/)(.*)"
+        groups = re.match(re_pw, engine_url)
+        engine_url_no_pw = "<unknown URL pattern>"
+        if len(groups.groups()) == 4 and len(groups.groups()[1]) > 0:
+            engine_url_no_pw = groups[1] + "***" + groups[3] + groups[4]
+        return cls(engine_url, engine_url_no_pw, **config)
```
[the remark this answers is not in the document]
windiana42: nice! `repr(sa.engine.make_url(engine_url))` does exactly what we want to do. Beware `str(make_url())` keeps password plain text.

src/pydiverse/pipedag/backend/table/sql.py
windiana42: It is now.

src/pydiverse/pipedag/backend/table/sql.py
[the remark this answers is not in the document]
NMAC427: Turns out there is also a [render_as_string](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.engine.URL.render_as_string) method.

src/pydiverse/pipedag/backend/table/sql.py
windiana42: Because it might be useful for further error messages.

pyproject.toml
windiana42: I think treating pydiverse-transform as an extra dependency comes closer to the loose coupling with the two libraries. You can run with or without the other. And it is easier to test both with having it as an extra.

pytest.ini
windiana42: I actually like it better if the pdtransform extension behaves similar than the mssql/db2 extras: you can install an environment with or without them and depending on this you expect some tests to succeed and not others. If you use importskip and importing fails also for -E pdtransform, then you won't see any error. I don't have a strong opinion on this though.

windiana42:
> As this mostly test that PydiverseTransformTableHook is implemented correctly, I propose that we create explicit tests for all TableHooks instead. Otherwise, there is nothing special about how pipedag treats transform.

I like to see more pydiverse.transform tests in pipedag. I just wanted to add one because I wanted to see whether it is working at all before trying that in a client environment.

NMAC427:
> I like to see more pydiverse.transform tests in pipedag. I just wanted to add one because I wanted to see whether it is working at all before trying that in a client environment.

Can you please elaborate on what exactly you are trying to test? Pydiverse pipedag and transform are independent projects. The only code in pipedag that has anything to do with transform is the `PydiverseTransformTableHook` implemented for the `SQLTableStore` and `DictTableStore`. As such, I don't see what else you could potentially want to test.

Because of this plugin system (thanks to the TableHooks), we can test the various components independently. If we know that `pydiverse.transform`, `PydiverseTransformTableHook`  and `pydiverse.pipedag` all work correctly independently, then we can also conclude that they work together.

pyproject.toml
NMAC427: In that case we would also have to create an extra for pandas, because it only gets used during the definition of the `PandasTableHook`. (I would even go as far as to suggest to remove pandas from the list of dependencies, as not to force any user of pipedag to use the specific version of pandas 'required' by pipedag).

QuantCo-mr-T:
There is not much code that needs testing. But there are some optional imports of pydiverse.transform and I wanted to test what exactly to put in auto-tables option of pipedag.yaml. And I like to see examples with pydiverse.transform that someone can easily copy to get started. The last aspect can be put in a separate repo though.

pyproject.toml
QuantCo-mr-T: Pandas and SQL alchemy are the base solutions for pydiverse.pipedag eager and lazy at least for now. And they might be reference points in terms of correctness forever.

QuantCo-mr-T: Interoperability of pandas and lazy code is one big feature of pipedag. So I think it is fine to define a minimum version and then keep track of compatibility with new versions from then on. We might test which is the oldest pandas version that pipedag runs fine with. But since pipedag will take months before reaching larger adoption and maybe >1year before adoption by not-contributors, this is not the most pressing issue.

windiana42:
> As this mostly test that `PydiverseTransformTableHook` is implemented correctly, I propose that we create explicit tests for all TableHooks instead.

Yes, we should add additional unit tests. But the "integration test" is better than nothing. And it also demos how to use it. This is not trivial with auto_tables. See following PRs which use the same technique for demonstrating ibis and polars integrations.

windiana42:
will merge without review to release pipedag 0.2.4 this week

Issue #39, opened 2023-01-24
Support for DB2 environment where CREATE SCHEMA is not permitted  


windiana42:
DB2 can be setup to create schemas whenever an object in that schema is created on demand. Strangely this works even without the permission to run CREATE SCHEMA explicitly. Thus we need to implement an option in pipedag.yaml for such an environment.

PR #40, opened 2023-01-24
Implemented #39: option avoid_drop_create_schema

windiana42:
Also removed a few unused imports and emphasized with a comment how instance_id parameterized tests are activated for a file/directory.

# Checklist


windiana42:
@NMAC427 this PR is actually the most urgent to merge to master and release as 0.2.2 since I already work with this change in a client environment to be able to work there. I know tests can be improved. We currently have just smoke tests in here. But in the client environment it will be extensively tested.

PR #44, opened 2023-01-28
Fix primary key / index with same column on different tables within schema

windiana42:
We have a bug that primary/index names are not separated per table in many database targets. This PR fixes it by adding table name to index names.

# Checklist


cmdevries (approved):
Makes sense to me 👍

PR #46, opened 2023-01-31
Speed up performance for DB2

windiana42:
This PR adds a lot of customization for DB2 targets which improve performance:
- lock on table level
- make primary key columns not nullable in the middle of CreateTableAsSelect compilation
- ALIAS instead of VIEWS
...

# Checklist

### DIFF
--- a/src/pydiverse/pipedag/backend/table/sql.py
+++ b/src/pydiverse/pipedag/backend/table/sql.py
@@ -1229,8 +1229,13 @@ class SQLAlchemyTableHook(TableHook[SQLTableStore]):
 
     @classmethod
     def retrieve(cls, store, table, stage_name, as_type):
+        table_name = table.name
+        if store.engine.dialect.name == "ibm_db_sa":
+            # DB2 has uppercase as case-insensitive default
+            if table_name.islower():
+                table_name = table_name.upper()
         return sa.Table(
-            table.name,
+            table_name,
             sa.MetaData(bind=store.engine),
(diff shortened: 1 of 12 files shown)


src/pydiverse/pipedag/backend/table/sql.py
```diff
@@ -736,6 +758,33 @@ def _commit_stage_read_views(self, stage: Stage):
...
+
+    @_create_read_views.dialect("ibm_db_sa")
+    def _create_read_views_ibm_db_sa(
+        self, conn, dest_schema, src_schema: Schema, src_tables: Iterable[sa.Table]
+    ):
+        # Instead of views create aliases which can be locked like tables and
+        # have primary keys
+        for table in src_tables:
+            self.execute(
+                CreateAlias(table.name, src_schema, table.name, dest_schema),
```
cmdevries, line 784: Is this the main source of the speedup?

cmdevries (approved):
Makes sense to me

src/pydiverse/pipedag/backend/table/sql.py
QuantCo-mr-T: This is at the heart of the speedup. We only create aliases while processing cache valid tasks. If we are 100% cache valid, we never copy the data and never swap Schemas / replace read views. When we encounter the first cache invalid task, we start copying the data for all aliases in parallel and on commit of schema, we replace aliases with the copy.

The aliases cannot stay after commit since their source may be dropped afterwards.

PR #54, opened 2023-02-09
Don't copy any data in case of 100% cache valid stages

windiana42:
Sounds like a nobrainer. But it is quite tricky to achieve since we only know half way through stage execution whether this is the case or not. So we start setting just aliases/read views and replace them by copying tables (in the background) when the first cache invalid task is discovered and the aliases are replaced before stage commit.

# Checklist 


windiana42:
Visible by test failures, there will have to be some more solidification work done. For DB2 and mssql the technique is already working in general. For Postgres it might be a bit more tricky since we can only place Read-Views as aliases. However, also for DB2/mssql we resolve those aliases before working with them (sqlalchemy can't work with them very well). The same can be done also for read views. They would just be placeholders in this case functioning a s a level of indirection.

.github/actions/test/action.yml
```diff
@@ -10,5 +10,5 @@ runs:
     - name: Test
       shell: bash
       run: |
-        poetry run pytest tests -ra ${DEBUG:+-s} ${{ inputs.arguments }}
+        poetry run pytest tests -ra -v -s ${DEBUG:+-s} ${{ inputs.arguments }}
```
windiana42, line 13: For now, I suggest to leave more verbose test output on CI. We are still having trouble with database deadlocks despite locking of stages.

.github/workflows/ci.yml
```diff
@@ -65,15 +65,133 @@ jobs:
         with:
           arguments: --workers 2
 
-  full_test:
+  full_test39:
```
windiana42, line 68: This lengthy duplication is not nice. We will try to undo this. One idea is strategy: max-parallel=1. However, this option did not work.

pyproject.toml
```diff
@@ -57,7 +57,7 @@ mssql = ["pyodbc", "pytsql"]
 ibm_db2 = ["ibm-db", "ibm-db-sa"]
 prefect = ["prefect", "docker-compose"]
 
-[tool.poetry.group.dev.dependencies]
+[tool.poetry.dev-dependencies]
```
windiana42, line 60: this seems to be the way for future poetry versions

src/pydiverse/pipedag/_typing.py
```diff
@@ -23,7 +23,7 @@ def decorator_hint(decorator: Callable) -> Callable:
 BaseMaterializable = Union[MPrimitives, MTypes]
 Materializable = Union[
     BaseMaterializable,
-    Dict[str, "Materializable"],
-    List["Materializable"],
-    Tuple["Materializable", ...],
+    dict[str, "Materializable"],
+    list["Materializable"],
+    tuple["Materializable", ...],
```
windiana42, line 28: This is an automated pyupdate change done by pre-commit hook.

src/pydiverse/pipedag/backend/table/sql.py
```diff
@@ -646,6 +651,13 @@ def _init_stage_schema_swap_mssql(self, stage: Stage):
                         conn=conn,
                     )
 
+            synonyms = self._get_mssql_sql_synonyms(transaction_schema.get())
+            for name in synonyms.keys():
+                self.execute(
+                    DropAlias(name, transaction_schema, if_exists=True),
```
windiana42, line 657: for mssql, we use synonyms as implementation of DropAlias/CreateAlias

src/pydiverse/pipedag/context/run_context.py
```diff
@@ -91,14 +95,24 @@ def __init__(self, flow: Flow):
 
         self.task_memo: defaultdict[Any, Any] = defaultdict(lambda: MemoState.NONE)
 
+        # deferred table store operations
+        self._thread_pool = ThreadPoolExecutor(
```
windiana42, line 99: we use multithreading here since all threads in process wait most of the time on I/O

windiana42 (commented):
Commented changes in PR

PR #60, opened 2023-04-14
Temp fix CI by running DB2 tests separately with only one concurrent configuration

windiana42:
It seems that the way DB2 service is run (priviledged service as opposed to docker-compose) causes trouble when running tests in parallel. Thus this CI configuration runs only one python version test on DB2 and runs the other tests without DB2.


windiana42:
Will merge without review since it just rearranges Ci test runs.

PR #61, opened 2023-04-15
closes #57: replace format_exception with format_exc

### LINKED ISSUE #57: Really support python 3.9
We currently use traceback.format_exception(exception) which only works since python 3.10. Tests did not catch it since this exception formatting probably is not actively checked by any test.

windiana42:
This improves compatibility for pythin 3.9.

# Checklist


windiana42:
Will merge without review since this is a bug fix dominated release.

PR #70, opened 2023-05-06
fix ibm_db_sa bug when copying dataframes from cache: uppercase table names by default

windiana42:
# Checklist

### DIFF
--- a/src/pydiverse/pipedag/backend/table/base.py
+++ b/src/pydiverse/pipedag/backend/table/base.py
@@ -17,6 +17,7 @@ from pydiverse.pipedag.materialize.metadata import (
 )
 from pydiverse.pipedag.materialize.util import compute_cache_key
 from pydiverse.pipedag.util import Disposable, requires
+from pydiverse.pipedag.util.naming import NameDisambiguator
 
 if TYPE_CHECKING:
     from pydiverse.pipedag import Stage
@@ -308,7 +309,9 @@ class BaseTableStore(Disposable, metaclass=_TableStoreMeta):
 
     # Dematerialize
 
-    def retrieve_table_obj(self, table: Table, as_type: type[T]) -> T:
+    def retrieve_table_obj(
+        self, table: Table, as_type: type[T], namer: NameDisambiguator | None = None
+    ) -> T:
         """Loads a table from the store
 
         Retrieves the table from the store, converts it to the correct
@@ -327,7 +330,7 @@ class BaseTableStore(Disposable, metaclass=_TableStoreMeta):
             )
 
         hook = self.get_r_table_hook(as_type)
-        return hook.retrieve(self, table, table.stage.current_name, as_type)
+        return hook.retrieve(self, table, table.stage.current_name, as_type, namer)
 
     # Metadata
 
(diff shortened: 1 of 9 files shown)


src/pydiverse/pipedag/backend/table/base.py
```diff
@@ -17,6 +17,7 @@
 )
 from pydiverse.pipedag.materialize.util import compute_cache_key
 from pydiverse.pipedag.util import Disposable, requires
+from pydiverse.pipedag.util.naming import NameDisambiguator
```
windiana42, line 20: NameDisambiguator / namer should help provide collision free short names to tables going into one task. This is used for setting aliases in sqlalchemy which make the query look much nicer.

src/pydiverse/pipedag/backend/table/sql.py
```diff
@@ -353,6 +353,17 @@ def execute(self, query, *, conn: sa.engine.Connection = None):
...
+    @engine_dispatch
+    def name_adj(self, name: str):
+        # some dialects need to replace all lowercase names by uppercase because
+        # they create uppercase table names by default
+        return name
+
+    @name_adj.dialect("ibm_db_sa")
+    def _name_adj_ibm_db_sa(self, name: str):
+        # DB2 creates tables uppercase if all lowercase given
+        return name.upper() if name.islower() else name
```
windiana42, line 365: We already had this concept in sql_ddl.py that IBM DB2 names needed to be converted from lowercase to uppercase if they are all lowercase. Now, this is available in sql.py via dialect specific implementations of self.name_adj(). see engine_dispatch.
The rational behind this is that all lowercase for us and for sqlalchemy means case insensitive identifiers. While postgres actually produces lowercase table/schema names from those, IBM DB2 produces all uppercase table/schema names.

src/pydiverse/pipedag/backend/table/sql.py
```diff
@@ -1451,8 +1478,8 @@ def materialize(
         task_info: TaskInfo,
     ):
         obj = table.obj
-        if isinstance(table.obj, sa.Table):
-            obj = sa.select(table.obj)
+        if isinstance(table.obj, (sa.Table, sa.sql.selectable.Alias)):
```
windiana42, line 1481: This is necessary since we don't get Tables into tasks any more but Alias objects.

src/pydiverse/pipedag/backend/table/sql.py
```diff
@@ -1451,8 +1478,8 @@ def materialize(
...
+            obj = sa.select("*").select_from(table.obj)
```
windiana42, line 1482: this produces nicer looking query for humans

src/pydiverse/pipedag/backend/table/sql.py
windiana42, line 1481: This is necessary since we don't get Tables into tasks any more but Alias objects.

tests/test_flows/test_sql_text_node.py
```diff
@@ -19,7 +19,11 @@ def table_1(script_path: str):
 
 @materialize(input_type=sa.Table, lazy=True)
 def table_2(script_path: str, dependent_table: Table):
-    sql = Path(script_path).read_text().replace("{{dependent}}", str(dependent_table))
+    sql = (
+        Path(script_path)
+        .read_text()
+        .replace("{{dependent}}", str(dependent_table.original))
```
windiana42, line 25: dependent_table is an Alias and thus dependent_table.original is the underlying table.

windiana42:
will merge without review to get release 0.3.0 out

PR #72, opened 2023-05-10
implemented ignore_task_version option

windiana42:
# Checklist


docs/reference/config.rst
```diff
@@ -806,16 +817,16 @@ example code for loading configuration
 .. code-block:: python
 
     flow = create_flow1()
-    flow.run()  # will internally run cfg=PipedagConfig.load().get(flow_name=flow.name)
+    flow.run()  # will internally run cfg=PipedagConfig.default.get(flow_name=flow.name)
```
windiana42, line 820: This is just a little documentation fix that should have been corrected long time ago

src/pydiverse/pipedag/materialize/core.py
```diff
@@ -76,7 +76,8 @@ class MaterializingTask(Task):
     :key version: The version of this task. Unless this task is lazy, you
         always have to bump / change the version number to ensure that
         the new implementation gets used. Else a cached result might be used
-        instead.
+        instead. version=None means the task is never cached. In addition,
```
NMAC427, line 79: I think `version=None means the task is never cached` could be worded more accurately. 

`version=None` in conjunction with `lazy=False` means that doesn't get cached (but the outputs still get materialized).

windiana42: Is fixed as requested with later PR (to avoid chain rebase)

Issue #51, opened 2023-02-08
Can we put an authentication token in communication between tasks and RunConfigServer?


windiana42:
The more functionality we offer by RunConfigServer, the more security relevant is the communication. Ideally, we hash the complete operation with a secret random seed determined at runtime. For this to make sense, the spawning of tasks which would include this seed must be secured as well. This would then be the responsibility of the Engine backend or framework we sit on top.

PR #83, opened 2023-06-02
Encrypt and authenticate IPC Server + Client communication

NMAC427:
Addresses PR #51 using [Fernet](https://cryptography.io/en/latest/fernet/#fernet-symmetric-encryption).


NMAC427:
Close #51

Issue #116, opened 2023-10-26
Simplify debugging of materialization write-back


windiana42:
At end of tasks, the final materialization may throw errors caused by database systems which are hard to foresee in pipedag user code. It would simplify debugging of such problems significantly if materialization can be triggered by the user directly during interactive debugging. It should also be possible to support automatic overwriting of tables to allow calling the materialization repeatedly even after it succeeded in one of the interactive test materialization calls. 

Databases may be limited in row-length or have trouble with one specific type constellation without reporting a good error message. Thus binary search is often the way to narrow down a problem. For dataframe writing, we could even automate this binary search for the column that breaks the materialization.

windiana42:
Another idea for this manual debugging materialization is optional checking of explicit dtypes. I.e. for String(n) columns: check that all strings to be written in this column have length <= n.

PR #117, opened 2023-11-03
Add a function to ad-hoc materialize a table for interactive debugging

nicolasmueller:
Add pipedag.debug.materialize_table, a function to ad-hoc materialize a table from a task for interactive debugging

# Checklist

Related issue: #116

### DIFF
--- a/src/pydiverse/pipedag/materialize/debug.py
+++ b/src/pydiverse/pipedag/materialize/debug.py
@@ -9,6 +9,7 @@ from pydiverse.pipedag.backend import SQLTableStore
 from pydiverse.pipedag.backend.table.sql.ddl import DropTable
 from pydiverse.pipedag.context import TaskContext
 from pydiverse.pipedag.materialize.core import MaterializingTask
+from pydiverse.pipedag.materialize.store import mangle_table_name
 from pydiverse.pipedag.util.hashing import stable_hash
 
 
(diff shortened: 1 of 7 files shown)


src/pydiverse/pipedag/materialize/debug.py
```diff
@@ -0,0 +1,80 @@
...
+    suffix = (
+        stable_hash(str(random.randbytes(8))) + "_0000"
+        if debug_suffix is None
+        else debug_suffix
+    )
+    old_table_name = table.name
+    if table.name is None:
+        table.name = task.name + "_" + suffix
+    elif table.name.endswith("%%"):
+        table.name = table.name[:-2] + suffix
```
windiana42, line 59: I think it would be nice to code share this name mangling with the original code that materializes tables. It will be more maintainable then.

src/pydiverse/pipedag/materialize/debug.py
```diff
@@ -0,0 +1,80 @@
...
+from pydiverse.pipedag.materialize.core import MaterializingTask
+from pydiverse.pipedag.util.hashing import stable_hash
+
+
+def materialize_table(
+    table: Table,
+    debug_suffix: str | None = None,
+    flag_task_debug_tainted: bool = True,
+    keep_table_name: bool = True,
+    replace_table: bool = True,
```
windiana42, line 20: The naming of `replace_table` might be optimized. Some dataframe materializations use if_exists="replace". This is a crude enum technique. However, people might be familiar with it. And we could raise an exception for all other string values.
Another alternative would be to rename it `drop_if_exists`. This would be closest to what this feature actually does and thus makes it easier to understand error messages reported by the database backend.

src/pydiverse/pipedag/materialize/debug.py
nicolasmueller: Changed in 971ee0545d5ffe59c23a69e3758003912d2b5995

PR #119, opened 2023-11-16
DB2 development on OS X arm64

nicolasmueller:
- Add documentation on how to develop on OS X with arm64 using a x86_64 toolchain for DB2.

- Add psycopg2-binary, which is needed for the OS X arm64 setup to pyproject.toml

# Checklist


windiana42:
On Linux, I receive the error:
```
Because pydiverse-pipedag depends on psycopg2-binary (>=2.9.3) which doesn't match any versions, version solving failed.
```

nicolasmueller:
> On Linux, I receive the error: [...]

Interestingly, the Github Actions test passed on Ubuntu.
Unfortunately, we cannot install psycopg2-binary via conda for OS X as the conda package is `noarch` (why is that for a binary package?).
I modified `pyproject.toml` to only depend on `psycopg2-binary` when running on OS X with `x86_64`. Judging by the output of the Github Actions tests, it seems that psycopg2-binary is no longer installed on Ubuntu.

windiana42 (approved):
looks good

PR #122, opened 2023-11-22
Materialization details

nicolasmueller:
# Checklist


pipedag.yaml
```diff
@@ -89,6 +89,7 @@ instances:
 
   ibm_db2:
     instance_id: pd_ibm_db2
+    compression: ["COMPRESS YES ADAPTIVE", "VALUE COMPRESSION"]
```
nicolasmueller, line 92: @windiana42  Is `pipedag.yaml` the right place to test this option? Should this be an instance-level option at all?

nicolasmueller:
@windiana42 Should specifying a compression method for a backend that does not support it result in an exception? Currently, we just print a warning.

src/pydiverse/pipedag/backend/table/sql/dialects/mssql.py
```diff
@@ -235,6 +235,12 @@ def _execute_materialize(
         if table.type_map:
             dtypes.update(table.type_map)
 
+        if table.compression:
+            store.logger.warning(
```
QuantCo-mr-T, line 239: I think this is worth a logger.error. For fail-fast we might even go for raising an exception (no need to implement now, just to give a feeling of error handling principles). You should not ask for a compression that is not supported by the dialect you are using.

src/pydiverse/pipedag/core/config.py
```diff
@@ -209,6 +209,7 @@ def get(
             network_interface=config["network_interface"],
             kroki_url=config.get("kroki_url"),
             attrs=Box(config["attrs"], frozen_box=True),
+            compression=config.get("compression"),
```
QuantCo-mr-T, line 212: Did you see config.rst for documenting configuration options?

QuantCo-mr-T (approved):
Minor change suggestions. Very nice implementation!

One could have gone for Pipedag enum to control compression Independent of Backend. But I think this is unrealistic. Compression Options are very dialect specific even if some common themes exist.

nicolasmueller:
@windiana42 Could you have a look at this again? Now, the `materialization_details` can be configured in `pipedag.yaml`. For now DB2 compression and table spaces are supported (not yet unlogged Postgres tables).

Some notes:

- I also had to add the `materialization_details` to the cached table metadata so that they are applied when a table is copied to a new schema.
- It is possible that you have to delete your DuckDB cache once, before using the new version.
- I had to hand craft the `CREATE TABLE` statement when writing a pandas dataframe to DB2 when using materialization details, since the table space cannot easily be changed after the creation (see `CreateTableWithSuffix`)

src/pydiverse/pipedag/backend/table/sql/dialects/mssql.py
nicolasmueller: In the newest version, there is a `logger.error` or an exception depending on strictness.

windiana42:
> I also had to add the materialization_details to the cached table metadata so that they are applied when a table is copied to a new schema.

We store a version number of our metadata in the metadata itself. Shall we bump this version number? Maybe we say we don't need it if we are backwards compatible with tables without materialization_details specified.

windiana42 (approved):
looks good

nicolasmueller:
> We store a version number of our metadata in the metadata itself. Shall we bump this version number? Maybe we say we don't need it if we are backwards compatible with tables without materialization_details specified.

It's probably better style to bump it, so I'll bump it. For DuckDB it seems necessary anyway to refresh the cache after this change.

src/pydiverse/pipedag/materialize/details.py
```diff
@@ -0,0 +1,105 @@
...
+    def from_dict(
+        cls: type[_T], d: dict[str, dict[str | list[str]]], strict: bool, logger
+    ) -> _T:
+        unsupported_arguments = set(d.keys()) - {
+            field.name for field in dataclasses.fields(cls)
+        }
+        if unsupported_arguments:
+            error_msg = (
+                f"The materialization arguments {unsupported_arguments} "
+                f"are not supported for DB2."
```
QuantCo-mr-T, line 32: I guess the DB2 reference is a relic of moving code around. It should not appear in generic code.

src/pydiverse/pipedag/backend/table/sql/sql.py
```diff
@@ -377,6 +407,22 @@ def execute(self, query, *, conn: sa.engine.Connection = None):
 
         return self._execute(query, conn)
 
+    def check_materialization_details_supported(self, label: str | None) -> None:
+        from pydiverse.pipedag.backend.table.sql.dialects import IBMDB2TableStore
```
QuantCo-mr-T, line 411: I guess this can also be removed. The parent class version always throws an error and some child classes may override the error throwing code with concrete functionality

QuantCo-mr-T (approved):
There are still two DB2 references in generic code

src/pydiverse/pipedag/backend/table/sql/sql.py
nicolasmueller: This is not there anymore in the newest version: https://github.com/pydiverse/pydiverse.pipedag/pull/125/files#diff-cbf7003cc01e080153ebf63516c520a1b21a40ebfd49c4a17b1ac019f748d34dL410

src/pydiverse/pipedag/materialize/details.py
nicolasmueller: Fixed in fa4f6ba

PR #128, opened 2023-12-08
Fix no materialization details

nicolasmueller:
This PR fixes a crash when providing DB2 with no materialization details. It also adds a `ibm_db2_materialization_details` instance and removes the materialization details from the `ibm_db2` instance.

# Checklist


windiana42 (approved):
looks good

PR #135, opened 2024-02-04
Use `polars-lts-cpu` instead of `polars` for x86_64 OS X

nicolasmueller:
When running pipedag on OS X with rosetta (i.e. `x86_64` emulation),  the polars import crashes since it assumes the existence of some CPU instructions not available in the emulator. Importing `polars-lts-cpu` instead in this case fixes the problem.

# Checklist


windiana42 (approved):
If it works for you on ARM MacOS, then this is ready to merge.

windiana42:
The poetry.lock file should already be checked by unit tests for Linux. I tested the change here on my machine once more including `poetry lock` and also found no issue.

PR #139, opened 2024-02-24
Enable the creation of external TableReferences

nicolasmueller:
Enable the creation of external TableReferences that reference tables in external schemas. This is useful to tell pipedag about tables that already exist in the database and should not be copied. It avoid manually reflecting them, which can sometimes be cumbersome as one might want to call a `Task` with a pipedag `Table` and sometimes with an external table.

# Checklist


NicolasMuellerQC:
Enabling external views for DuckDB depends on https://github.com/duckdb/duckdb/issues/10322

windiana42:
> Enabling external views for DuckDB depends on https://github.com/duckdb/duckdb/issues/10322

We could implement a workaround since this problem is quite common among some SQLAlchemy backends. We have such a workaround in other places of the codebase.

src/pydiverse/pipedag/backend/table/sql/sql.py
```diff
@@ -1278,48 +1280,49 @@ class TableReference:
...
-    To get the correct schema name, you must use undocumented and non-public
-    parts of the pipedag API. To help you get started, we'll provide you with
-    this example, however, be warned that this might break without notice::
+    You can use a `TableReference` to tell pipedag about a table that exists
+    in an external schema::
 
         @materialize(version="1.0")
         def task():
-            from pydiverse.pipedag.context import ConfigContext, TaskContext
+            return Table(TableReference("external_schema"), "name_of_table")
```
windiana42, line 1302: We might rename TableReference since it is tied to external and we might forbid the use of the name != None in dag.Table.
```suggestion
            return Table(ExternalTableReference("name_of_table", schema="external_schema"))
```

windiana42 (changes requested):

nicolasmueller:
>We could implement a workaround since this problem is quite common among some SQLAlchemy backends. We have such a workaround in other places of the codebase.

Done.

src/pydiverse/pipedag/backend/table/sql/hooks.py
```diff
@@ -57,11 +57,11 @@ def materialize(
         source_tables = [
             dict(
                 name=tbl.name,
-                schema=store.get_schema(tbl.stage.current_name).get(),
+                schema=store.get_schema(tbl.stage.current_name, tbl).get(),
```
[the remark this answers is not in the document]
nicolasmueller: A DB2 test has been added in a5769bc701d276cf346e86d17526ffd408313890

NicolasMuellerQC:
@windiana42 Let me know if it looks good now 😃

src/pydiverse/pipedag/backend/table/sql/ddl.py
```diff
@@ -706,6 +706,7 @@ def visit_create_table_as_select_ibm_db_sa(create: CreateTableAsSelect, compiler
         src_tables = [
             f"{preparer.format_schema(tbl['schema'])}.{preparer.quote(tbl['name'])}"
             for tbl in create.source_tables
+            if tbl["shared_lock_allowed"]
```
windiana42, line 709: Just to remind myself: this is needed to prevent DB2 dialect from locking external table reference input tables if not allowed.

windiana42 (approved):
Looks good to me (I changed a few things directly on the branch).

src/pydiverse/pipedag/util/json.py
```diff
@@ -117,7 +117,7 @@ def get_stage(name: str):
         tbl.stage = get_stage(d["stage"])
         tbl.cache_key = d["cache_key"]
         tbl.external_schema = d.get("external_schema")
-        tbl.shared_lock_allowed = d.get("shared_lock_allowed", True)
+        tbl.shared_lock_allowed = d.get("shared_lock_allowed", False)
```
nicolasmueller, line 120: @windiana42 
This would change the default behavior for existing DB2 pipedag instances even for non-external tables (even though the shared locks are not currently working due to the bug fixed in #144 ) to not lock source tables (since the metadata `shared_lock_allowed` is not available there yet). 

I think the default here should be `True` as it is here:

https://github.com/pydiverse/pydiverse.pipedag/blob/ae1fd3614933972f6971699a107a2b7739624b3e/src/pydiverse/pipedag/materialize/container.py#L60

windiana42: Correct. I only wanted the default to be False for ExternalTableReferences.

windiana42:
```suggestion
        tbl.shared_lock_allowed = d.get("shared_lock_allowed", True)
```

windiana42:
@nicolasmueller good catch! did you find something else?

NicolasMuellerQC:
>@nicolasmueller good catch! did you find something else?

@windiana42  No, I am fine with merging then 👍

PR #140, opened 2024-02-24
Make lazy task output tables which do not provide a query string cache invalid for subsequent reading tasks

nicolasmueller:
The cache validity of a lazy task cannot be judged when the tables output by the task do not provide query strings. This should result in the task being cache invalid which it currently does not. This task fixes this.

# Checklist

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.6.10 (2024-XX-XX)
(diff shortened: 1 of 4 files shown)


NicolasMuellerQC:
@windiana42  This should be ready now :)

windiana42 (approved):
I suggest adding an issue for hashing eager tables as their hashed content (should only be executed for lazy=True). This PR looks good.

nicolasmueller:
> I suggest adding an issue for hashing eager tables as their hashed content (should only be executed for lazy=True). This PR looks good.

Tracked in #141

Issue #142, opened 2024-02-28
Enable PipedagConfig construction from dictionary


windiana42:
If someone likes to programmatically produce pipedag configuration, it should be possible to construct a PipedagConfig object from a dictionary. Ideally, it is possible to input PipedagConfig.raw_config into that constructor and yield exactly the same PipedagConfig.config_dict.

An extended version of implementing this feature would offer some programmatic calls which can produce config dictionaries which are good default configurations with very little code.

Hacky workaround code as long as this feature is not available:
```python
    pipedag_config = PipedagConfig()
    pipedag_config.raw_config["table_store_connections"][connection_name]["args"]["url"] = override_engine_url
    pipedag_config.config_dict = pipedag_config._PipedagConfig__parse_config(
        pipedag_config.raw_config
    )
    cfg = pipedag_config.get(instance)
```

windiana42:
Ideally this would allow for much simpler getting started documentation with programmatically created duckdb configuration:

- https://github.com/Quantco/vectorization-tutorial/blob/main/pipedag.yaml
- https://github.com/pydiverse/pydiverse.pipedag/blob/main/docs/source/quickstart.md
- https://github.com/pydiverse/pydiverse.pipedag/blob/main/docs/package/README.md
- https://github.com/pydiverse/pydiverse.pipedag/blob/main/README.md

windiana42:
Closed with #142

PR #148, opened 2024-03-07
#142: Easier configuration setup 

windiana42:
Currently, the only way to configure pipedag is to provide a configuration yaml. More alternatives including configuration in code should be provided.

# Checklist


windiana42:
@NMAC427 FYI: we make examples using code based configuration and duckdb as a containerless database more prominent in documentation.

NMAC427:
Seems reasonable.
Though I would probably rename `get_basic_pipedag_config` to `create_basic_pipedag_config`.

windiana42:
Since this PR already touched example code and configuration widely, I also included a switch from ZookeeperLockManager to DatabaseLockManager for default tests. DatabaseLockManager is the one I would recommend to everyone who is working with a database target that supports it (mssql, postgres, db2).

PR #152, opened 2024-03-10
Fix duckdb example documentation

windiana42:
# Checklist


windiana42:
@NMAC427 @nicolasmueller I will fix this documentation hickup right away for readthedocs (push to main is sufficient). But I won't take this as a reason to make a new release. The documentation on pypi is wrong though until until 0.7.1 is released for other reasons.

windiana42:
I tested both pip and conda-forge installation techniques and now it works for me at least. CI tested it previously as well, but there was a pipedag.yaml in the discovery path which wasn't supposed to be used.

PR #155, opened 2024-03-15
Adding max_nchars_query_print option for MSSqlTableStore

DominikZuercherQC:
Added the `max_nchars_query_print` option for the `MSSqlTableStore` allowing configuration of the maximum number of characters in a SQL query that are printed to the log

### DIFF
--- a/src/pydiverse/pipedag/backend/table/sql/dialects/mssql.py
+++ b/src/pydiverse/pipedag/backend/table/sql/dialects/mssql.py
@@ -51,6 +51,10 @@ class MSSqlTableStore(SQLTableStore):
         reach across, but it is not guaranteed to be identical to scripts executed
         by a SQL UI.
 
+    :param max_nchars_query_print:
+        Maximum number of characters in a SQL query that are printed when
+        logging the query before the log is truncated. Defaults to 5000 characters.
+
     .. _Microsoft SQL Server:
         https://www.microsoft.com/en/sql-server/
     .. _pytsql.executes:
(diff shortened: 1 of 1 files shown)


src/pydiverse/pipedag/backend/table/sql/dialects/mssql.py
```diff
@@ -51,6 +51,10 @@ class MSSqlTableStore(SQLTableStore):
         reach across, but it is not guaranteed to be identical to scripts executed
         by a SQL UI.
 
+    :param max_nchars_query_print:
```
windiana42, line 54: I would prefer a bit more human name: max_query_print_length
Don't have a strong opinion though.

src/pydiverse/pipedag/backend/table/sql/dialects/mssql.py
```diff
@@ -64,10 +68,12 @@ def __init__(
         *args,
         disable_pytsql: bool = False,
         pytsql_isolate_top_level_statements: bool = True,
+        max_nchars_query_print: int = 5000,
```
windiana42, line 71: I would actually default to 200000 or even 500000.

windiana42 (approved):
Generally, nice and clean solution!

src/pydiverse/pipedag/backend/table/sql/dialects/mssql.py
DominikZuercherQC: Updated it

src/pydiverse/pipedag/backend/table/sql/dialects/mssql.py
DominikZuercherQC: Agree, increased it

nicolasmueller:
Do you want to add a `changelog.md` entry :)  @DominikZuercherQC ?

windiana42:
Yes, I think this is worth a changelog.md entry. I will just push it to another branch...
@DominikZuercherQC The PR template comment actually adds a checkbox that reminds to add a changelog entry. It seems you removed this part when creating this PR.

PR #157, opened 2024-03-15
Fix duplicate table bug

nicolasmueller:
Fix bug where a table could not be created if an external table with the same name is loaded in the same stage.

# Checklist

- [x] Add test to cover this

### DIFF
--- a/src/pydiverse/pipedag/context/run_context.py
+++ b/src/pydiverse/pipedag/context/run_context.py
@@ -637,7 +637,8 @@ class RunContext(BaseContext):
         response = self._request(
             "add_names",
             stage.id,
-            [t.get_run_context_name() for t in tables],
+            # ExternalTableReferences don't create a table in schema of stage
+            [t.name for t in tables if t.external_schema is None],
             [b.name for b in blobs],
         )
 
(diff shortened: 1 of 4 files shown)


src/pydiverse/pipedag/context/run_context.py
```diff
@@ -650,7 +650,7 @@ def remove_names(self, stage: Stage, tables: list[Table], blobs: list[Blob]):
         self._request(
             "remove_names",
             stage.id,
-            [t.name for t in tables],
+            [t.get_run_context_name() for t in tables],
```
windiana42, line 653: It does not matter if we reference an external table multiple times within a stage:
```suggestion
            # ExternalTableReferences don't create a table in schema
            [t.name for t in tables if t.external_schema is None],
```

src/pydiverse/pipedag/context/run_context.py
```diff
@@ -637,7 +637,7 @@ def add_names(
         response = self._request(
             "add_names",
             stage.id,
-            [t.name for t in tables],
+            [t.get_run_context_name() for t in tables],
```
windiana42, line 640:
```suggestion
            # ExternalTableReferences don't create a table in schema
            [t.name for t in tables if t.external_schema is None],
```

windiana42 (changes requested):
ExternalTableReferences should not appear in the name check at all. It is allowed to have the same external table referenced multiple times within one schema.

windiana42 (approved):
This looks really nice and clean. Thank you!

PR #158, opened 2024-03-18
Refactor DB2 dependency in generic code.

windiana42:
Polymorphism is used instead.

# Checklist

### DIFF
--- a/src/pydiverse/pipedag/backend/table/sql/dialects/ibm_db2.py
+++ b/src/pydiverse/pipedag/backend/table/sql/dialects/ibm_db2.py
@@ -183,9 +183,7 @@ class IBMDB2TableStore(SQLTableStore):
             for st in ("data", "index", "long")
         }
 
-    def get_create_table_suffix(
-        self, materialization_details_label: str | None
-    ) -> str | None:
+    def get_create_table_suffix(self, materialization_details_label: str | None) -> str:
         table_spaces = self._get_table_spaces(materialization_details_label)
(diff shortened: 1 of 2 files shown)


windiana42:
@nicolasmueller I prefer the polymorphism solution to getting rid of the DB2 dependency in generic code

nicolasmueller (changes requested):
Currently the signature of `get_create_table_suffix` is inconsistent:
The generic table takes `table: Table` while the DB2 version takes `get_create_table_suffix: str | None`.
So either we have to adjust the generic version or fold `resolve_materialization_details_label` into the DB2 version.

I'd be in favor of the first version since the function is used without `resolve_materialization_details_label` in `copy_raw_sql_tables_to_transaction`.

windiana42:
> Currently the signature of `get_create_table_suffix` is inconsistent: The generic table takes `table: Table` while the DB2 version takes `get_create_table_suffix: str | None`. So either we have to adjust the generic version or fold `resolve_materialization_details_label` into the DB2 version.
> 
> I'd be in favor of the first version since the function is used without `resolve_materialization_details_label` in `copy_raw_sql_tables_to_transaction`.

Oops. Yes the idea to simplify usage by just providing a table object is more tricky. So I went back to the old interface. One could consider to have take `Table | Stage` . I would think this is nicer encapsulation. But not nice enough to refactor it now.

NicolasMuellerQC (approved):
Looks good now!

PR #162, opened 2024-03-20
Fix #127 regarding UNLOGGED postgres tables during copy table.

### LINKED ISSUE #127: Support for Postgres unlogged tables when copying tables
Currently `CopyTable` does not yet support Postgres unlogged tables. Support should be added for unlogged tables to be created consistently when the `unlogged` `materialization_detail` is set.

windiana42:
# Checklist


windiana42:
The first test suggests that #127 might be partially solved (for non-RAW-SQL scripts) by #160 refactoring

PR #166, opened 2024-03-20
Add option to `Flow.run` to force execution of all tasks

### LINKED ISSUE #165: Disable cache validation for a specific run
It should be possible to force the execution of all tasks when running a flow.

This is useful to invalidate the cache manually, when the user has more up-to-date information than the cache function.

nicolasmueller:
Fixes #165 

Further changes:
- [x] Make `ignore_cache_function=True` prevent cache function calls
- [x] Make the new `force_task_execution=True` option in `flow.run()` imply `ignore_cache_function=True`
- [x] Materialize lazy tasks when `config_context._force_task_execution=True`.

# Checklist

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -6,7 +6,9 @@
 - Added max_query_print_length parameter to MSSqlTableStore to limit the length of the printed SQL queries.
     Default is max_query_print_length=500000 characters.
 - Fix bug when creating a table with the same name as a `Table` given by `ExternalTableReference` in the same stage.
-- Add support for `force_task_execution` in `Flow.run` to force execution of tasks even if they are cache valid.
+- Add support for `force_task_execution` in `Flow.run` to force execution of tasks even if they are cache valid. This implies `ignore_cache_function=True`.
+- Make `ignore_cache_function=True` prevent cache function calls.
+- Materialize lazy tasks, when they are executed without stage context.
 
 ## 0.7.1 (2024-03-11)
 - Fix bug when Reading DECIMAL(precision, scale) columns to pandas task (precision was interpreted like for Float where 
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -5,8 +5,8 @@
     Function create_basic_pipedag_config() just has a kroki_url parameter which defaults to None.
 - Added max_query_print_length parameter to MSSqlTableStore to limit the length of the printed SQL queries.
     Default is max_query_print_length=500000 characters.
-- Fix bug when creating a table with the same name as a `Table` given by `ExternalTableReference` in the same stage  
-
+- Fix bug when creating a table with the same name as a `Table` given by `ExternalTableReference` in the same stage.
+- Add support for `force_task_execution` in `Flow.run` to force execution of tasks even if they are cache valid.
 
 ## 0.7.1 (2024-03-11)
 - Fix bug when Reading DECIMAL(precision, scale) columns to pandas task (precision was interpreted like for Float where 
--- a/src/pydiverse/pipedag/backend/table/base.py
+++ b/src/pydiverse/pipedag/backend/table/base.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING, Any, Generic
 import structlog
 from typing_extensions import Self
 
+from pydiverse.pipedag import ConfigContext
 from pydiverse.pipedag._typing import T, TableHookResolverT
 from pydiverse.pipedag.context import RunContext, TaskContext
 from pydiverse.pipedag.errors import CacheError
@@ -334,7 +335,7 @@ class BaseTableStore(TableHookResolver, Disposable):
 
         Used when `lazy = True` is set for a materializing task.
         """
-
+        config_context = ConfigContext.get()
         try:
             hook = self.get_m_table_hook(type(table.obj))
             query_str = hook.lazy_query_str(self, table.obj)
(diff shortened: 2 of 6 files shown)


nicolasmueller:
This PR also fixes an issue where a lazy task was not executed even when run standalone (without a stage) if it was cache valid.

NMAC427:
If I have time, I’d like to review this a bit more closely. (Everything cache related is prime bug territory)

src/pydiverse/pipedag/backend/table/base.py
```diff
@@ -357,6 +358,9 @@ def store_table_lazy(
             metadata = self.retrieve_lazy_table_metadata(
                 query_hash, task_cache_info.cache_key, table.stage
             )
+            if config_context._force_task_execution:
+                self.logger.info("Forced task execution")
+                raise CacheError("Forced task execution")
```
windiana42, line 363: wouldn't it make sense to do this at beginning of try block?

docs/source/changelog.md
```diff
@@ -5,8 +5,10 @@
     Function create_basic_pipedag_config() just has a kroki_url parameter which defaults to None.
 - Added max_query_print_length parameter to MSSqlTableStore to limit the length of the printed SQL queries.
     Default is max_query_print_length=500000 characters.
-- Fix bug when creating a table with the same name as a `Table` given by `ExternalTableReference` in the same stage  
-
+- Fix bug when creating a table with the same name as a `Table` given by `ExternalTableReference` in the same stage.
+- Add support for `force_task_execution` in `Flow.run` to force execution of tasks even if they are cache valid. This implies `ignore_cache_function=True`.
+- Make `ignore_cache_function=True` prevent cache function calls.
+- Materialize lazy tasks, when they are executed without stage context.
```
windiana42, line 11:
```suggestion
- Materialize lazy tasks, when they are executed without flow/stage context.
```

src/pydiverse/pipedag/materialize/core.py
```diff
@@ -500,7 +500,7 @@ def __call__(self, *args, **kwargs):
         input_hash = stable_hash("INPUT", input_json)
 
         cache_fn_hash = ""
-        if task.cache is not None:
+        if task.cache is not None and not config_context.ignore_cache_function:
```
windiana42, line 503: This has a downside. Previously, we would still know cache_fn_hash of the last valid run of all source tasks. Thus if you then disable ignore_cache_function, it still would not require rerunning the graph if there is really no change on source inputs. With this change, whenever a source node was invalidated with ignore_cache_function=True, it will have to rerun on setting ignore_cache_function=False. This is relevant for example for a scenario where we would wrap pipedag into a pipeline UI like Vertex AI. There, we would expose individual stages separately but with ignore_cache_function=True from 2nd stage onwards. This change would render automatic cache invalidation useless over multiple runs. 
=> it might be necessary to enable both behaviors. ignore_cache_function with and without recording cache_fn_hash.

windiana42: Proposal for names of the options needed: ignore_cache_function + disable_cache_function + force_task_execution
disable_cache_function would imply ignore_cache_function but not vice versa.

Btw. Originally, ignore_cache_function was called ignore_fresh_input. It was renamed because the new name is closer to what the option does. But conceptionally, the old name was better since it named the desired effect.

src/pydiverse/pipedag/materialize/core.py
```diff
@@ -559,7 +559,7 @@ def __call__(self, *args, **kwargs):
 
                     try:
                         _, cache_metadata = store.retrieve_cached_output(
-                            task, input_hash, cache_fn_hash
+                            task, input_hash, ""
```
windiana42, line 562:
```suggestion
                            task, input_hash, cache_fn_hash=""
```

windiana42 (approved):
I actually wonder how the previously existing force_task execution worked. It seemed there was the idea to make it a config option but this was not fully enacted. I actually would highly appreciate if force_task_execution would also become a config option that can be controlled on instance level. In this case it is questionable whether it actually makes sense to offer an override parameter to the Flow.run() method.

I currently don't understand the relation of these two snippets in the previously existing code:
```python
        # Evolve config using the arguments passed to flow.run
        config = config.evolve(
            fail_fast=(fail_fast if fail_fast is not None else config.fail_fast),
            ignore_cache_function=ignore_cache_function,
            force_task_execution=(
                # If subflow consists of a subset of tasks (-> not a subset of stages)
                # then we want to skip cache validity checking to ensure the tasks
                # always get executed.
                subflow.is_tasks_subflow
            ),
        )
```
and
```python
    _force_task_execution: bool = False
```
I could not find any code that copies force_task_execution to _force_task_execution.
```

src/pydiverse/pipedag/materialize/core.py
windiana42: We should also document alongside that caching is also disabled in case the cache metadata has the wrong version number. This can be fixed by removing metadata with the command line tool pipedag-manage. We probably should mention metadata version numbers per release in changelog.md in a way that makes it apparent in case of which upgrades, pipedag-manage needs to be called to keep caching active.

NicolasMuellerQC:
> I could not find any code that copies force_task_execution to _force_task_execution.

I believe that happens in the `ConfigContext.evolve` function calls this: https://www.attrs.org/en/stable/api.html#attrs.evolve. It says e.g.
> private attributes should be specified without the leading underscore, just like in __init__.

windiana42:
To make it even more complicated, we should not forget this feature when thinking about refactoring config options:

> cfg = ConfigContext.get().evolve(ignore_task_version=True)

PR #177, opened 2024-04-02
Added test triggering [from] bug

DominikZuercherQC:
# Checklist

### DIFF
--- a/tests/test_sql_ddl.py
+++ b/tests/test_sql_ddl.py
@@ -17,6 +17,7 @@ def test_insert_into():
         "Select (SELECT (SELECT 1 FROM Z) FROM Y) From X": (
             "Select (SELECT (SELECT 1 FROM Z) FROM Y) INTO a.b From X"
         ),
+        "Select a.[from] from a": "Select a.[from] from a",
     }
     for raw_query, expected_query in test_pairs.items():
         res = insert_into_in_query(raw_query, "a", "b")


src/pydiverse/pipedag/backend/table/sql/ddl.py
```diff
@@ -778,6 +778,14 @@ def insert_into_in_query(select_sql, schema, table):
         regex = re.compile(r"\b" + marker + r"\b", re.IGNORECASE)
         for match in regex.finditer(select_sql):
             match_start = match.span()[0]
+            match_end = match.span()[1]
+            # special case handling: ignore escaped from
+            prev_char = select_sql[match_start - 1]
+            next_char = select_sql[match_end]
```
windiana42, line 784: probably noone likes spaces in column names. But technically, it would be nice to support this situation as well:
```sql
DROP TABLE IF EXISTS _tbl01;
select 1 as [ from ] INTO _tbl01;

select '"'+COLUMN_NAME+'"', * from master.INFORMATION_SCHEMA.COLUMNS where TABLE_NAME='_tbl01'
```

src/pydiverse/pipedag/backend/table/sql/ddl.py
```diff
@@ -778,6 +778,14 @@ def insert_into_in_query(select_sql, schema, table):
...
+            if (prev_char == "[" and next_char == "]") or (
+                prev_char == '"' and next_char == '"'
+            ):
+                break
             prev = select_sql[0:match_start]
             # ignore marker in subqueries in select columns
             if prev.count("(") == prev.count(")"):
```
windiana42, line 791: How about catching it with this logic:
`if prev.count("(") == prev.count(")") and prev.count("[") == prev.count("]") and prev.count('"') % 2 == 0`

We cannot simply look for whitespaces between FROM and quoting characters:
```sql
DROP TABLE IF EXISTS _tbl01;
select 1 as [.from] INTO _tbl01;
select '"'+COLUMN_NAME+'"', * from master.INFORMATION_SCHEMA.COLUMNS where TABLE_NAME='_tbl01'
```

DominikZuercherQC: Makes sense! Actually the fix I proposed did not work as I intended as I misunderstood what the function was supposed to do :)

Issue #181, opened 2024-04-09
Add validation tasks to a Stage which get access to multiple schemas at once (last committed schema of this instance or another instance for this Stage)


windiana42:
Before committing a stage, due to schema swapping, we are in a good position to compare the current prepared schema with the last committed stage version of this pipeline instance or another pipeline instance.

Interface options are numerous. It could look like this:
```python
@stage_validator(input_type=sa.Table)
def my_validator(tbls_transaction_schema: dict[str, sa.Alias], tbls_compare_schema: dict[str, sa.Alias]):
  # has access to autoloaded tables of both schemas
  return dag.Table(compare_result_table, name="_cmp_%%")

with Stage("stage_2"):
    lazy_2 = lazy_task_2(lazy_1, b)
    lazy_3 = lazy_task_3(lazy_2)
    eager = eager_task(lazy_1, b)
    cmp3 = my_validator(lazy_3)  # we could limit the dictionaries to only include only tables referenced in arguments
    post_cmp3 = post_process(cmp3)  # this could be a normal task
    my_validator()  # by default, all tables of both schemas are reflected so the validator could also report about which tables are missing or unexpected
```

We can define that in case of `@stage_validator(input_type=None)`, the decorated function only receives schema names as values in the dictionary:
```python
def get_store():
  return ConfigContext.get().store.table_store

@stage_validator(input_type=None)
def my_validator(tbls_transaction_schema: dict[str, str], tbls_compare_schema: dict[str, str]):
  store = get_store()
  for table_name in tbls_compare_schema.keys():
    schema1 = tbls_transaction_schema[table_name]
    schema2 = tbls_compare_schema[table_name]
    assert store.execute(f"""SELECT (SELECT count(*) FROM "{schema1}"."{table_name}") - (SELECT count(*) FROM "{schema2}"."{table_name}") as diff""").fetchone()[0] == 0
```

Some users might even want to prevent the screening of tables in schema altogether. In case the user provides explicit table names, they could still be provided as dict[table_name, schema_name]. However, when no tables are provided, we would simply just have the schema names:
```python
@stage_validator(input_type=None, disable_reflect_schema=True)
def my_validator(tbls_transaction_schema: str, tbls_compare_schema: str):
```

It is easy to define syntactic sugar for engine access in user space:
```python
def get_engine():
  return ConfigContext.get().store.table_store.engine
```

[...]

PR #182, opened 2024-04-12
Add group nodes for structuring flows, as task ordering barrier, and for clusting tasks in visualization

windiana42:
# Checklist


windiana42:
This PR introduces baseline testing for visualizations: https://github.com/pydiverse/pydiverse.pipedag/blob/group_nodes/tests/util/baseline.py
It is used here: https://github.com/pydiverse/pydiverse.pipedag/blob/group_nodes/tests/test_run_group_node.py#L190

This means it stores a copy of the visualization URLs of flows in the repo and compares the actual output to that baseline in the repo. In case of changes a `test_run_group_node.updated.json` file is created but needs to be copied manually to `test_run_group_node.json` to accept the changes.

windiana42:
> What exactly is the purpose of the barrier nodes?

There are cases where it is easier to explicitly ask for tasks being executed in the order of declaration irrespective of explicit argument based dependencies. The idea was actually born while working on #181. A common usecase there is to declare a stage validator task which is executed at the end of preparing a stage. Thus the associated decorator `@input_stage_versions` could be made execute after all `@materialize` tasks within the stage unless explicit dependencies demand a different ordering. However the `@input_stage_versions` decorator can also be used to copy filtered input into a stage. Thus, in rare cases, one might want to execute it first. This can be done with such a barrier. Among current pipedag users, the idea that task order can be controlled more explicitly was very well received and also demanded in some unrelated discussions.

windiana42:
@NMAC427 @NicolasMuellerQC @DominikZuercherQC I plan to merge this PR early tomorrow in order to release 0.9.0.

PR #196, opened 2024-06-03
feat(): add upload_table and download_table functions to PandasTableHook

DominikZuercherQC:
Add an easy way to customize upload and download behavior of pandas and polars tables to/from the table store

### DIFF
--- a/tests/test_table_hooks/test_pandas.py
+++ b/tests/test_table_hooks/test_pandas.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import datetime as dt
 import string
+from typing import Any
 
 import pandas as pd
 import pyarrow as pa
@@ -11,6 +12,8 @@ from packaging.version import Version
 
 import tests.util.tasks_library as m
 from pydiverse.pipedag import *
+from pydiverse.pipedag.backend.table.sql.hooks import PandasTableHook
+from pydiverse.pipedag.backend.table.util import DType
 
 # Parameterize all tests in this file with several instance_id configurations
 from tests.fixtures.instances import DATABASE_INSTANCES, with_instances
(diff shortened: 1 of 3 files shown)


tests/test_table_hooks/test_pandas.py
```diff
@@ -326,3 +329,78 @@ def global_df_task(df):
         should_swap_inputs = True
         f.run()
         in_tables_spy.assert_called(2)
+
+
+@with_instances("postgres")
+class TestPandasCustomHook:
+    def test_custom_upload(self):
+        @ConfigContext.get().store.table_store.register_table(pd)
```
QuantCo-mr-T, line 337: @DominikZuercherQC did you check that it is guaranteed that the derived class hook always takes precedence over the parent class hook? From the back of my head, all registered hooks are asked whether they can (de)materialize a table and the first one responding yes is taken.

An optional Parameter might help where you can specify which Hook should be replaced by the newly registered one. This should be remembered so import in any order should be allowed.

QuantCo-mr-T (approved):

tests/test_table_hooks/test_pandas.py
DominikZuercherQC: > @DominikZuercherQC did you check that it is guaranteed that the derived class hook always takes precedence over the parent class hook? 

`TableHookResolver.__registered_tables() `goes through the registered hooks in inverse order, so this should actually be guaranteed. 
I find this behavior quite intuitive. Do we need to add a warning / flag here?

tests/test_table_hooks/test_pandas.py
```diff
@@ -326,3 +329,78 @@ def global_df_task(df):
...
+                "int8": pd.array(Values.INT8, dtype="Int8"),
+            }
+        )
+
+        @materialize()
+        def numpy_input():
+            return df
+
+        @materialize(input_type=pd.DataFrame)
+        def verify_cutom(t):
```
NicolasMuellerQC, line 363: @DominikZuercherQC Should this be `verify_custom`? :)

PR #197, opened 2024-06-06
fix(): add previous_hook_replace argument for register_table

DominikZuercherQC:
- Added `replace_hooks` argument for `register_table`. This avoids reliance on import order.
Given hooks are removed from the table store for which the new hook is registered. If the replace_hooks are located in parent classes, they have lower priority anyways. Please, note, that replace_hooks argument is not effective in case the replacement hooks was registered on a subclass of the table store where the new hook is added. 
If `replace_hooks=None`, the new hook is just added to the list of available hooks.

- Added another hook for download of polars Tables (FYI @nicolasmueller )

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -1,10 +1,11 @@
 # Changelog
 
-## 0.9.3 ()
+## 0.9.3 (2024-06-11)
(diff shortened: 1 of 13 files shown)


windiana42 (approved):
Improves reliability of hook replacement tests.

PR #201, opened 2024-06-18
feat(): add truncation of primary keys and indexes

DominikZuercherQC:
MSSQL does not allow identifiers longer than 128 characters. 
Similarly, postgres automatically truncates identifiers at 63 characters.
This PR automatically truncates all primary key and index identifiers at 63 characters.

# Checklist

### DIFF
--- a/src/pydiverse/pipedag/backend/table/sql/ddl.py
+++ b/src/pydiverse/pipedag/backend/table/sql/ddl.py
@@ -42,6 +42,11 @@ from sqlalchemy.sql.type_api import TypeEngine
 
 from pydiverse.pipedag import Schema
 
+# Postgres truncates identifiers at 63 characters
+# MSSQL does not allow identifiers longer than 128 characters
+MAX_LENGTH_PK = 63
+MAX_LENGTH_INDEX = 63
+
 
 class CreateSchema(DDLElement):
     def __init__(self, schema: Schema, if_not_exists=False):
@@ -265,6 +270,20 @@ class DropFunction(DDLElement):
         self.if_exists = if_exists
 
 
+def truncate_key(prefix, table_name, columns, max_length):
+    pre = prefix + table_name.lower() + "_"
+    pk = pre + columns
+    if len(pk) > max_length:
+        col_hash = str(hash(table_name.lower() + columns))
+        col_hash_trunc = col_hash[: max_length - len(pre)]
+        pk = pre + col_hash_trunc
+    if len(pk) > max_length:
+        pk = (
+            prefix + str(hash(table_name.lower() + columns))[: max_length - len(prefix)]
+        )
+    return pk
+
+
 class AddPrimaryKey(DDLElement):
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


src/pydiverse/pipedag/backend/table/sql/ddl.py
```diff
@@ -265,6 +270,20 @@ def __init__(self, name, schema: Schema | str, if_exists=False):
...
+def truncate_key(prefix, table_name, columns, max_length):
+    pre = prefix + table_name.lower() + "_"
+    pk = pre + columns
+    if len(pk) > max_length:
+        col_hash = str(hash(table_name.lower() + columns))
+        col_hash_trunc = col_hash[: max_length - len(pre)]
+        pk = pre + col_hash_trunc
+    if len(pk) > max_length:
+        pk = (
+            prefix + str(hash(table_name.lower() + columns))[: max_length - len(prefix)]
```
windiana42, line 282: I think this is dangerous for collisions. I would reserve a certain number of characters for a hash. Then you can truncate the table name and append the hash here.

src/pydiverse/pipedag/backend/table/sql/ddl.py
```diff
@@ -265,6 +270,20 @@ def __init__(self, name, schema: Schema | str, if_exists=False):
         self.if_exists = if_exists
 
 
+def truncate_key(prefix, table_name, columns, max_length):
+    pre = prefix + table_name.lower() + "_"
+    pk = pre + columns
+    if len(pk) > max_length:
+        col_hash = str(hash(table_name.lower() + columns))
```
windiana42, line 277: It might be important to use a stable hash here. hash() changes from run to run. We have a utility function called stable_hash in pipedag.

windiana42 (changes requested):

src/pydiverse/pipedag/backend/table/sql/ddl.py
DominikZuercherQC: Thanks I was not aware that hash() is not constant over runs. Using stable_hash() now.

src/pydiverse/pipedag/backend/table/sql/ddl.py
DominikZuercherQC: I changed the logic a bit such that the hash is used first

src/pydiverse/pipedag/backend/table/sql/ddl.py
```diff
@@ -41,6 +41,12 @@
 from sqlalchemy.sql.type_api import TypeEngine
 
 from pydiverse.pipedag import Schema
+from pydiverse.pipedag.util.hashing import stable_hash
+
+# Postgres truncates identifiers at 63 characters
+# MSSQL does not allow identifiers longer than 128 characters
+MAX_LENGTH_PK = 63
+MAX_LENGTH_INDEX = 63
```
windiana42, line 49: It would be nice to have an issue that talks about making this configurable as argument to the SQLTableStore. This could even be labeled "good first issue".

windiana42 (approved):
Nice!

PR #202, opened 2024-06-18
feat(): Added ignore_position_hash option for SQLTableStore

DominikZuercherQC:
Added `ignore_position_hashes` option to `SQLTableStore`. 
  If `True`, the position hashes of tasks are not checked when retrieving the inputs of a task from the cache.
  This can prevent caching errors when evaluating subgraphs. 
  For this to work a task may never be used more than once per stage.

# Checklist

### DIFF
--- a/src/pydiverse/pipedag/backend/table/dict.py
+++ b/src/pydiverse/pipedag/backend/table/dict.py
@@ -97,13 +97,13 @@ class DictTableStore(BaseTableStore):
                 f"'{task.name}' with cache key '{task.input_hash}', yet"
             ) from None
 
-    def retrieve_all_task_metadata(self, task: MaterializingTask) -> list[TaskMetadata]:
+    def retrieve_all_task_metadata(self, task: MaterializingTask, ignore_position_hashes: bool) -> list[TaskMetadata]:
         task_metadata = []
         for m in (
             *self.metadata[task.stage].values(),
             *self.t_metadata[task.stage].values(),
         ):
-            if m.name == task.name and m.position_hash == task.position_hash:
+            if m.name == task.name and ((m.position_hash == task.position_hash) or ignore_position_hashes):
                 task_metadata.append(m)
         return task_metadata
 
--- a/src/pydiverse/pipedag/backend/table/dict.py
+++ b/src/pydiverse/pipedag/backend/table/dict.py
@@ -98,7 +98,7 @@ class DictTableStore(BaseTableStore):
             ) from None
 
     def retrieve_all_task_metadata(
-        self, task: MaterializingTask, ignore_position_hashes: bool
+        self, task: MaterializingTask, ignore_position_hashes: bool = False
     ) -> list[TaskMetadata]:
         task_metadata = []
         for m in (
--- a/src/pydiverse/pipedag/backend/table/dict.py
+++ b/src/pydiverse/pipedag/backend/table/dict.py
@@ -97,13 +97,17 @@ class DictTableStore(BaseTableStore):
                 f"'{task.name}' with cache key '{task.input_hash}', yet"
             ) from None
 
-    def retrieve_all_task_metadata(self, task: MaterializingTask, ignore_position_hashes: bool) -> list[TaskMetadata]:
+    def retrieve_all_task_metadata(
+        self, task: MaterializingTask, ignore_position_hashes: bool
+    ) -> list[TaskMetadata]:
         task_metadata = []
         for m in (
             *self.metadata[task.stage].values(),
             *self.t_metadata[task.stage].values(),
         ):
-            if m.name == task.name and ((m.position_hash == task.position_hash) or ignore_position_hashes):
+            if m.name == task.name and (
+                (m.position_hash == task.position_hash) or ignore_position_hashes
+            ):
                 task_metadata.append(m)
         return task_metadata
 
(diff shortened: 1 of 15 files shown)


tests/test_run_subflow.py
```diff
@@ -295,3 +295,18 @@ def test_run_specific_stage_with_table_blob():
...
+        with Stage("subflow_s1"):
+            x1 = m.create_tuple(1, 1)
+            x2 = m.create_tuple(x1, x1)
+
+        with Stage("subflow_s2") as s2:
+            _ = m.create_tuple(x1, x2)
+
+    f.run()
+    # check that retrieval from cache works
+    f.run(s2)
```
windiana42, line 312: Is this test really testing what we are interested in? Wouldn't we need to create different flows with modified sources of the tasks  producing x1 and x2?

I am actually not sure it is ok to call the same task multiple times with disable_position_hash=true. Outputs of those tasks might be mixed up on cache retrieval.

src/pydiverse/pipedag/backend/table/sql/sql.py
```diff
@@ -1389,18 +1396,23 @@ def retrieve_task_metadata(
         )
 
     def retrieve_all_task_metadata(self, task: MaterializingTask) -> list[TaskMetadata]:
+        if self.ignore_position_hashes:
+            match_condition = sa.and_(
+                self.tasks_table.c.name == task.name,
+                self.tasks_table.c.stage == task.stage.name,
+            )
+        else:
+            match_condition = sa.and_(
```
windiana42, line 1405: how about removing the `if` in line 1399 and doing here in the `else` branch:
```
if not self.ignore_position_hashes:
  match_condition = match_condition & (self.tasks_table.c.position_hash == task.position_hash)
```

src/pydiverse/pipedag/backend/table/sql/sql.py
```diff
@@ -1389,18 +1396,23 @@ def retrieve_task_metadata(
         )
 
     def retrieve_all_task_metadata(self, task: MaterializingTask) -> list[TaskMetadata]:
```
windiana42, line 1398: See comment on PR: I don't think SQLTableStore is the right place for configuring disable_position_hash. Thus, I it should be a parameter to this function.

Top level in Flow.run(), the parameter is also not generally disabling position hashes. It is disabling position hashes for retrieving inputs when running subflows. I saw that the code is also modifying the call of get_output_from_store on table references. Probably, the use of position hashes should also be manually choosable there.

windiana42 (commented):
Configuration of disable_position_hash feature should be somewhere else than SQLTableStore. It is limited to running subflows and retrieving data from table references. This should be reflected in the configuration option name or documentation. The test should work with modified flow.

DominikZuercherQC:
> Do you really think disable_position_hash is a feature that people like to configure per instance?

No, I agree that this is probably more of a global choice. So no need to make this configurable per instance.

> Ignoring position hash is a dangerous feature that can lead to invalid cache recovery. Thus, I rather see it as a debugging mode and a big warning should inform the user that he is at his own risk. Would it work to only enable it as option to Flow.run()? It mainly makes sense in combination with running subflows which are also configured there.

In our case the position_hash regularly causes issues when trying to rerun subflows. So I would actually like to disable it permanently in our pipeline just to make development easier. I do not see that much potential for danger here as long as one does not use two tasks with the same name in a stage (could we not even add a check for this actually?) or am I missing something.
However, I would be totally fine to only have this as an option for Flow.run().

src/pydiverse/pipedag/backend/table/sql/sql.py
DominikZuercherQC: added

tests/test_run_subflow.py
DominikZuercherQC: Agree this was basically only a smoke test so far. 
I now modified the test such that it checks that the flag actually fixes subflow evaluation

src/pydiverse/pipedag/backend/table/dict.py
```diff
@@ -97,13 +97,17 @@ def retrieve_task_metadata(
                 f"'{task.name}' with cache key '{task.input_hash}', yet"
             ) from None
 
-    def retrieve_all_task_metadata(self, task: MaterializingTask) -> list[TaskMetadata]:
+    def retrieve_all_task_metadata(
+        self, task: MaterializingTask, ignore_position_hashes: bool
```
windiana42, line 101: It must be clear that ignore_position_hashes=true is a dangerous operation that should only be done if you know what you are doing. This can either be solved by a doc string, or by a logging message that gives a warning inside this function. I would actually prefer a default argument ignore_position_hashes=false to make clear that this is "normal".

windiana42 (approved):
I like the configuration via Flow.run(). There may be more configuration options provided in the future, but this is a good start.

src/pydiverse/pipedag/backend/table/dict.py
DominikZuercherQC: @windiana42 I set all ignore_position_hashes options to False by default.
I also added a warning for flow.run(). I would like to avoid setting warnings at lower levels as we would use this feature extensively and I am afraid that it clutters up the logs

windiana42 (approved):

src/pydiverse/pipedag/core/flow.py
```diff
@@ -322,7 +333,12 @@ def run(
             flow.run(task_1, task_4)
 
         """
-
+        self.logger.warn(
```
windiana42, line 336: @DominikZuercherQC This warning is unconditional and should only be printed for ignore_position_hashes==True.

PR #203, opened 2024-07-02
bug(): Complex tasks using imperative materialization break caching + release `v 0.9.4`

DominikZuercherQC:
@windiana42 My expectation is that even when using imperative materialization in a task it should stay cache valid if its inputs did not change. In complex cases this can break.

This PR serves as a minimal example:` test_materialize.py::test_imperative_minimal_example`

### DIFF
--- a/src/pydiverse/pipedag/context/trace_hook.py
+++ b/src/pydiverse/pipedag/context/trace_hook.py
@@ -1,0 +1,87 @@
+from __future__ import annotations
+from typing import TYPE_CHECKING
+
+import structlog
+
+if TYPE_CHECKING:
+    from pydiverse.pipedag import Result, Task, Table
+    from pydiverse.pipedag.context import RunContextServer, RunContext
+    from pydiverse.pipedag._typing import Materializable
+    from pydiverse.pipedag.materialize.cache import TaskCacheInfo
+    from pydiverse.pipedag.materialize.metadata import TaskMetadata
+
+
+class TraceHook:
+    def run_init_context_server(self, run_context_server: RunContextServer):
+        pass
+
+    def run_init_context(self, run_context: RunContext):
+        pass
+
+    def run_complete(self, result: Result):
+        pass
+
+    def stage_init(self, stage_id: int, success: bool):
+        pass
+
+    def stage_pre_commit(self, stage_id: int):
+        pass
+
+    def stage_post_commit(self, stage_id: int, success: bool):
+        pass
+
+    def task_cache_status(self, task: Task, input_hash: str, cache_fn_hash: str, cache_metadata: TaskMetadata | None = None, cached_output: Materializable | None = None, cache_valid: bool | None=None, lazy: bool | None=None):
+        pass
+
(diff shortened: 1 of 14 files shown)


windiana42:
I can confirm the expectation. I need to debug to find the reason.

windiana42:
It takes a bit longer to fix because of implementing some groundwork for searching similar caching issues. The problem seems already found in the order of Table.assumed_dependencies:
```
    [<Table '_temp_res2_m3qtk77lrs6vpirivro5_0001' (stage)>, <Table '_temp_res2_m3qtk77lrs6vpirivro5_0000' (stage)>]
```
vs.
```
    [<Table '_temp_res2_m3qtk77lrs6vpirivro5_0000' (stage)>, <Table '_temp_res2_m3qtk77lrs6vpirivro5_0001' (stage)>]
```

src/pydiverse/pipedag/context/trace_hook.py
```diff
@@ -0,0 +1,152 @@
...
+        input_hash: str,
+        cache_fn_hash: str,
+        cache_metadata: TaskMetadata | None = None,
+        cached_output: Materializable | None = None,
+        cache_valid: bool | None = None,
+        lazy: bool | None = None,
+    ):
+        pass
+
+    def task_complete(self):
```
AaronTackeQC, line 46: I love this, this will get rid of all the dirty log-sniffing I had to do for estimating execution times.

To make it event better: 
- Why isn't the task passed here? 
- And could we get a `task_started` for exact timing?

Generally, I'd prefer a consistent naming scheme like `pre_flow`, `post_flow`, `pre_stage` `post_stage`/`pre_stage_commit`, `post_stage_commit`, `pre_task`, `post_task` - haven't thought long and hard about shortcomings, though.

windiana42: @AaronTackeQC I think some refinement of the names and parameters are still needed. Feel free to just push changes in this branch. I will leave it open until next week.

We will certainly have another PR elaborating on the TraceHook concept. It should be parametrizable via configuration. Some time tracing functionality would be good to always provide via the result object. Then comes a bit more complex logic to store history of run times per cache validity state in the database. And the user still might want something utterly different. In the end, it might useful to have a list of TraceHook classes called one after another in order to separate concerns.

windiana42: Naming is non-trivial. I like to distinguish different groups by first part of name. There is run level, stage level, task level, query level, and cache transfer level. A lot of pre and post information is redundant. task_post_call is identical to task_pre_materialize. The task_cache_status point coincides with mostly task_pre_dematerialize, but it is not exactly the same and the parameters are tied to cache status.
On the run level, for run_post_xxx it is hard to chose xxx. Thus, I prefer run_complete. Going with run_complete, I like task_complete better than task_post_materialize especially because task_post_dematerialize is a lot more cumbersome than task_pre_call.

windiana42 (approved):
Looks ready to merge for me.

PR #212, opened 2024-07-31
Fix polars >= 1

nicolasmueller:
Currently retrieving a polars dataframe with polars 1.2 crashes. This is because we currently use `read_database` instead of `read_database_uri`, but hand it a URI. My guess is that polars 1 got stricter with it's behavior.

# Checklist


tests/test_table_hooks/test_polars.py
```diff
@@ -186,7 +186,7 @@ def download_table(
             connection_uri: str,
         ):
             # FYI: this fails for duckdb and thus we fall back to pandas hook
-            df = pl.read_database(query, connection_uri)
+            df = pl.read_database_uri(query, connection_uri)
```
QuantCo-mr-T, line 189: Are we sure there is a Polars version for python 3.9 which offers the read_database_uri function?

QuantCo-mr-T (approved):

tests/test_table_hooks/test_polars.py
QuantCo-mr-T: Potentially we have to change the Polars version dependency constraint to ensure the read_database_uri function exists

windiana42: We still need to require polars >= 0.19. But this should be fine since it support python 3.9: https://prefix.dev/channels/conda-forge/packages/polars

windiana42 (changes requested):
Please update dependencies to 'polars >= 0.19'

windiana42 (approved):

PR #216, opened 2024-08-17
feat(): Task links

DominikZuercherQC:
Idea for a new feature:
Add the option to pass `task_links` to `flow.run()`.
`task_links` defines a mapping from tasks to (external) tables in the database.
When the flow is executed all tasks run as usual expect for the tasks listed in the mapping. These tasks are not executed, instead an `ExternaTableReference` is established for them.
If the option is not used `flow.run()` works as usual. 

Usecase: Development on a task requires testing it on a large dataset. Development in the original database is discouraged as it is used by many users. Duplicating / rerunning the entire database takes too long / too many computational resources.
-> This feature allows to use the tables in the original database without overhead, only run the tasks/stages interesting for the development at hand and leaves the other users undisturbed.

Note: Most of the changes actually originate from moving `container.py` to another submodule which was needed to solve a circular import (maybe there is a better way?)

@windiana42 / @nicolasmueller WDYT?

# Checklist


windiana42:
I would prefer if the syntax of this PR would be similar to: https://github.com/pydiverse/pydiverse.pipedag/issues/217

This might look as follows:
```python
    with Flow() as f:
        with Stage("sql_table_origin"):
            _ = make_external_table()

        with Stage("sql_table_linked"):
            table = duplicate_table_reference()
            _ = consume(table)
    external_link = table

    with StageLockContext():
        # Normal execution. duplicate_table_reference should return an empty table
        result = f.run()
        assert result.get(table, as_type=pd.DataFrame).shape[0] == 0

    with StageLockContext():
        # Linked execution. Body of duplicate_table_reference should not be executed,
        # instead output is referenced from the linked table
        result = f.run(
            inputs={
                external_link: ExternalTableReference(
                    "external_table",
                    schema="external_schema",
                )
            }
        )
```

windiana42:
We could even support a feature for the usecase to reference the same table in another pipeline instance:
```python
   cfg = PipedagConfig.default.get(instance)
   input_cfg = PipedagConfig.default.get("full") 

    with Flow() as f:
        with Stage("sql_table_origin"):
            _ = make_external_table()

        with Stage("sql_table_linked"):
            table = duplicate_table_reference()
            _ = consume(table)
    external_link = table

    with StageLockContext():
        # Linked execution. Body of duplicate_table_reference should not be executed,
        # instead output is referenced from the linked table
        result = f.run(
            inputs={
                external_link: input_cfg
            },
            config=cfg  # still optional
        )
```
This syntax would even allow to dematerialize dicts/lists/constants and multiple tables that appear in the output of duplicate_table_reference from the cache of "full" instance.

docs/source/examples/raw_sql.md
```diff
@@ -25,7 +25,7 @@ import sqlalchemy as sa
 
 from pydiverse.pipedag import Flow, Stage, materialize
 from pydiverse.pipedag.context import ConfigContext, StageLockContext
-from pydiverse.pipedag.materialize.container import RawSql
+from pydiverse.pipedag.container.container import RawSql
```
windiana42, line 28: What happened here with module names?

windiana42: I also had problems with circular dependencies around container in the past. However, "container.container" just looks odd.

src/pydiverse/pipedag/engine/sequential.py
```diff
@@ -34,6 +40,7 @@ def run(self, flow: Subflow, ignore_position_hashes: bool = False, **run_kwargs)
                             run_context=run_context,
                             config_context=config_context,
                             ignore_position_hashes=ignore_position_hashes,
+                            override=inputs.get(task) if inputs else None,
```
windiana42, line 43: Can't we replace the inputs in row 36 rather than overriding a task? I actually wonder where TaskGetItem is lazily resolved. It might be inside results[in_t].

We could even implement an algorithm to avoid calculating all tasks that are part of a subgraph with tasks that are only feeding into references mentioned in inputs. Sketch of idea (not tested!):
```python
input_tasks = {t.task if isinstance(t, TaskGetItem) else t for t in inputs.keys()}
skip_candidates = {task for flow.get_tasks() if task in inputs_tasks}
last_added = skip_candidates
while len(last_added) > 0:
  last_added = {task for task in added_task.input_tasks.values() for added_task in last_added}
  skip_candidates.update(last_added)

last_removed = skip_candidates
while len(last_removed) > 0:
  last_removed = skip_candidates & {task for task in active_task.input_tasks.values() for active_task in flow.get_tasks() if active_task is not in skip_candidates}
  skip_candidates -= last_removed  
```

windiana42 (changes requested):
Sorry for the lengthy review. I think it would be good to not replace a task but communication links between tasks when defining inputs.

src/pydiverse/pipedag/container/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+    Table,
+)
+
+__all__ = [
+    "Table",
+    "Blob",
+    "RawSql",
+    "ExternalTableReference",
+    "Schema",
+]
```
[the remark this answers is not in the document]
DominikZuercherQC: Makes sense. Moved content to `__init__.py`

src/pydiverse/pipedag/engine/prefect.py
```diff
@@ -176,8 +186,15 @@ def pipedag_flow():
 
         return pipedag_flow
 
-    def run(self, flow: Subflow, ignore_position_hashes: bool = False, **kwargs):
+    def run(
+        self,
+        flow: Subflow,
+        ignore_position_hashes: bool = False,
+        task_links=None,
```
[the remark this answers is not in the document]
DominikZuercherQC: ups, thanks

src/pydiverse/pipedag/engine/dask.py
```diff
@@ -45,7 +45,18 @@ def __init__(self, **dask_compute_kwargs):
...
+        flow: Subflow,
+        ignore_position_hashes: bool = False,
+        inputs=None,
+        **run_kwargs,
+    ):
+        _ = ignore_position_hashes
+        if inputs:
+            raise NotImplementedError(
+                "The inputs argument is currently not supported for the Dask engine."
+            )
```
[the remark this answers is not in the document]
DominikZuercherQC: Agree. I added it for Prefect and Dask as well

docs/source/examples/raw_sql.md
DominikZuercherQC: moved `container.py` contents to `__init__.py`

windiana42 (approved):
Looks much better!

PR #222, opened 2024-09-05
fix(): bugfix for flow.run(inputs)

DominikZuercherQC:
The inputs feature did not work if the depended tasks were not previously executed.
This PR should fix this

# Checklist

### DIFF
--- a/src/pydiverse/pipedag/engine/base.py
+++ b/src/pydiverse/pipedag/engine/base.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any
 
-from pydiverse.pipedag import ExternalTableReference, RawSql, Table, Task
+from pydiverse.pipedag import ExternalTableReference, Table, Task
 from pydiverse.pipedag.core.task import TaskGetItem
 from pydiverse.pipedag.util import Disposable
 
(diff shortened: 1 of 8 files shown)


src/pydiverse/pipedag/engine/base.py
```diff
@@ -45,30 +45,6 @@ def _replace_task_inputs_with_const_inputs(
     task_inputs: dict[int, Any], inputs: dict[tuple[int, int | str | None], Table]
```
windiana42, line 45: Shouldn't inputs be renamed inputs_resolved, here, because the key is a tuple?

tests/test_inputs.py
```diff
@@ -65,9 +65,11 @@ def identity(table: pd.DataFrame):
     with StageLockContext():
         # Linked execution. Body of duplicate_table_reference should not be executed,
         # instead output is referenced from the linked table
+        task_to_link = f["sql_table_linked"].get_task("duplicate_table_reference")
```
windiana42, line 68: I would expect that this could be replaced by a reference to a link in the wiring. If it is a TaskGetItem, then one could simply take the task member of it. But it is fine to leave it as is.

tests/test_inputs.py
```diff
@@ -65,9 +65,11 @@ def identity(table: pd.DataFrame):
     with StageLockContext():
         # Linked execution. Body of duplicate_table_reference should not be executed,
```
windiana42, line 66: I doubt the current implementation prevents any execution of tasks. It only ignores the outputs they produce when calling consumers.

windiana42 (approved):

src/pydiverse/pipedag/engine/base.py
DominikZuercherQC: Yes I have to fix the type hints

tests/test_inputs.py
DominikZuercherQC: True, the comment is incorrect.

PR #228, opened 2024-10-03
fix: Use isinstance to support subclasses of values for deepmap

btat-qc:
For example this enables support for NamedTuples, making return types more clear from pipedag tasks

# Checklist

### DIFF
--- a/src/pydiverse/pipedag/util/deep_map.py
+++ b/src/pydiverse/pipedag/util/deep_map.py
@@ -3,6 +3,7 @@
 Heavily inspired by the builtin copy module of python:
 https://github.com/python/cpython/blob/main/Lib/copy.py
 """
+
 from __future__ import annotations
 
 from typing import Callable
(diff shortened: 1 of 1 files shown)


src/pydiverse/pipedag/util/deep_map.py
```diff
@@ -19,13 +20,11 @@ def deep_map(x, fn: Callable, memo=None):
     if y is not _nil:
         return y
 
-    cls = type(x)
-
-    if cls == list:
+    if isinstance(x, list):
         y = _deep_map_list(x, fn, memo)
-    elif cls == tuple:
+    elif isinstance(x, tuple):
```
windiana42, line 25: @btat-qc we talked about extensions like this in deep_map before. The question is what is what the expected behavior should be. `deep_map(v, lambda x: x)` should be very close to a deep copy without changing anything. With the change you proposed, the NamedTuple would be converted to a tuple if I am not mistaken.
I would suggest to explicitly handle NamedTuples and data classes here to guarantee that they are deep copied to the same type.

windiana42 (changes requested):
I am not sure your goal is achieved by this change. The NamedTuple is converted to a tuple and loses its names.

We actually have `deep_map` as a pretty fundamental tool in both pipedag and transform which means it should work quite consistently over a very large code base according to general assumptions. I would expect `deep_map(v, lambda x: x)` to behave very close to deep_copy for collections (non-collection elements are not copied).

I suggest to explicitly handle NamedTuple and python data-classes to produce exactly the same type after the copy than it was before.

src/pydiverse/pipedag/util/deep_map.py
btat-qc: I see - that makes sense! Thanks I'll think about how to solve this and add some tests

PR #234, opened 2025-02-05
Fix DuckDB 1.1 incompatibility

Revan1017:
There is a speedup mechanism for writing dataframes to duckdb in pipedag which requires access to a raw database connection of duckdb. Previously, this raw connection was opened fresh extracting the duckdb URL from the sqlalchemy engine. Since DuckDB 1.1, this results in an error, because the raw connection is setup differently than the sqlalchemy connection. This PR proposes a fix to get the raw_connection() from the SQLAlchemy engine directly. The raw_connection is used to reference a dataframe from the python scope directly: `INSERT... SELECT * FROM df` where `df` is the variable holding the dataframe  

# Checklist

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.9.9 (2025-02-dd)
+- Fix incompatibility with DuckDB 1.1.
+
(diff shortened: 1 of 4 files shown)


windiana42 (approved):
Nice and short fix.

PR #237, opened 2025-02-07
pipedag compatibility with pydiverse.transform >=0.2

Revan1017:
This PR makes pydiverse.pipedag compatible with pydiverse.transform >0.2.0 while still maintaining support for pydiverse.transform <0.2.0.
However pydiverse.pipedag still doesn't support pydiverse.transform 0.2.0, as it significasntly differs from every other verison in files and functions
Changes:
- Added a new Tablehook in hooks.py for pydiverse.transform >0.2.0
- Added a new Tablehook in parquet.py and a Version Detector for pydiverse.transform >0.2.0
- Detecting Version (old/new) by testing imports (try-except)
- Added a NotImplementedError("pydiverse.transform 0.2.0 isn't supported"), when pydiverse.transform 0.2.0 is tried to run
- Updated README.md to inform about incompatibility with pydiverse.transform 0.2.0
- Updated tests/test_table_hooks/test_pdtransform to also have the version-detection mechanism in order to test correctly
- Pinned prefect version down due to incompatibility
- Dropped python 3.9 due to pdt not supporting it anymore
- Removed ibm-on-osx support because of problems with pixi.lock and pixi.toml

# Checklist


.github/workflows/nightly_tests.yml
```diff
@@ -58,7 +58,6 @@ jobs:
           - py312all
           - py311all
           - py310all
-          - py39all
```
verticube, line 61: could you please add to the PR description why we decided to drop Python 3.9 from the tests?

pixi.toml
```diff
@@ -37,7 +37,7 @@ kazoo = ">=2.8.0"
 dask = ">=2022.1.0"
 
 [feature.prefect.dependencies]
-prefect = ">=2.13.5"
+prefect = ">=2.13.5, <3.0.0"
```
verticube, line 40: and could you please also add a note that/why we pinned the prefect versions?

pixi.toml
```diff
@@ -68,15 +67,21 @@ arrow-odbc = ">=7.0.4"
...
 # IBM DB2
 [feature.ibm-db]
 # ibm-db is not available on Apple Silicon
 # https://ibm-data-and-ai.ideas.ibm.com/ideas/DB2CON-I-92
-platforms = ["linux-64", "osx-64", "win-64"]
-[feature.ibm-db.target.osx.pypi-dependencies]
-# ibm_db is not on conda-forge for macOS
-# https://github.com/ibmdb/db2drivers/issues/3
-ibm-db = ">=3.1.4"
+platforms = ["linux-64", "win-64"]  # TODO "osx-arm64"
```
verticube, line 80: ...and that/why we removed ibm-on-osx support?

verticube (changes requested):
lgtm, please add some pointers to the PR description.

I'd appreciate a second opinion from others, e.g., @nicolasmueller

pixi.toml
verticube: @pavelzw maybe that's also something you could elaborate on. I'm wondering if we should maybe unbundle this particular change. wdyt?

pavelzw: The issue was that ibm-db on macos didn't have a conda-forge build back then (https://github.com/conda-forge/ibm_db-feedstock/pull/77) and just got one recently in https://github.com/conda-forge/ibm_db-feedstock/pull/79. In a future PR (#231) i will re-add macos support for ibm-db with the new package.

This lead to us using the ibm-db build on pypi for macos (in combination with telling people to install gcc next to clang) and thus activating `pypi-dependencies` in pixi.
This introduces some issues, namely in some cases (haven't found a minimal reproducer for this issue here, though) where pixi thinks the lockfile is out of date (which leads to `pixi install --locked` failing). xref https://github.com/prefix-dev/pixi/issues/1295, https://github.com/prefix-dev/pixi/issues/1417

> I'm wondering if we should maybe unbundle this particular change

this would require a change of the lockfile which will likely lead to ci breakages. if it's not high effort to fix these breakages (or introduce corresponding pinnings in pixi.toml), we could do it 🤷🏻

src/pydiverse/pipedag/backend/table/cache/parquet.py
```diff
@@ -190,12 +190,37 @@ def retrieve(
...
 except ImportError:
     pdt = None
+    pdt_old = None
+    pdt_new = None
 
 
-@ParquetTableCache.register_table(pdt)
-class PydiverseTransformTableHook(TableHook[ParquetTableCache]):
+@ParquetTableCache.register_table(pdt_old)
+class PydiverseTransformTableHookOld(TableHook[ParquetTableCache]):
```
DominikZuercherQC, line 223: Is there no more indicative naming possible here instead of just Old/New? 😅

src/pydiverse/pipedag/backend/table/sql/hooks.py
```diff
@@ -880,6 +908,100 @@ def lazy_query_str(cls, store, obj: pdt.Table) -> str:
...
+        table = table.copy_without_obj()
+        # if isinstance(t._impl, Polars):
+        #     table.obj = t >> collect()
+        #     hook = store.get_hook_subclass(PandasTableHook)
+        #     return hook.materialize(store, table, stage_name)
+        try:
+            table.obj = sa.text(t >> build_query())
+            hook = store.get_hook_subclass(SQLAlchemyTableHook)
+            return hook.materialize(store, table, stage_name)
+        except Exception:
```
DominikZuercherQC, line 951: Can we capture a more specific exception here instead of the general Exception?

windiana42 (approved):
Needs a bit of cleanup (osx ibm-db2) but can go to main branch.

src/pydiverse/pipedag/backend/table/cache/parquet.py
windiana42: The Old part will be retired at some point just not yet.

.github/workflows/nightly_tests.yml
windiana42: Retiring python 3.9 is fine by now.

PR #254, opened 2025-07-13
Make pydot optional dependency

windiana42:
# Checklist

### DIFF
--- a/src/pydiverse/pipedag/core/flow.py
+++ b/src/pydiverse/pipedag/core/flow.py
@@ -3,12 +3,12 @@
 
 import base64
 import random
+import types
 from collections import defaultdict
 from collections.abc import Callable, Iterable
 from typing import TYPE_CHECKING
 
 import networkx as nx
-import pydot
 import structlog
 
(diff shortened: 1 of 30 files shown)


src/pydiverse/pipedag/core/flow.py
```diff
@@ -524,6 +531,8 @@ def visualize(self, result: Result | None = None, visualization_tag: str | None
         return dot
 
     def visualize_url(self, result: Result | None = None, visualization_tag: str | None = None) -> str:
+        if pydot.Dot is None:
+            return "<please install pydot to receive a visualization URL>"
```
pavelzw, line 535: shouldn't we raise a `RuntimeError` here as well?

pavelzw: then should we maybe return `str | None` instead?

windiana42:
The problem is that the caller cannot convert the None to a good message
for the user.

We could go with the Exception and then have the automatic call catch it
and convert the message to a string that does not sound like an
error/exception.

PR #268, opened 2025-08-14
feat: Add `write_local_table_cache` to `Result.get()`

nicolasmueller:
Make it possible to enable writing to the local table cache after the flow has finished when calling `Result.get()`.

# Checklist

### DIFF
--- a/src/pydiverse/pipedag/materialize/core.py
+++ b/src/pydiverse/pipedag/materialize/core.py
@@ -725,7 +725,7 @@ def _get_output_from_store(
     root_task = task if isinstance(task, Task) else task.task
 
     store = ConfigContext.get().store
-    with DematerializeRunContext(root_task.flow):
+    with DematerializeRunContext(root_task.flow, allow_write_local_table_cache=False):
         cached_output, _ = store.retrieve_most_recent_task_output_from_cache(
             root_task, ignore_position_hashes=ignore_position_hashes
         )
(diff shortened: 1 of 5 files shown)


src/pydiverse/pipedag/materialize/core.py
```diff
@@ -725,7 +725,7 @@ def _get_output_from_store(
     root_task = task if isinstance(task, Task) else task.task
 
     store = ConfigContext.get().store
-    with DematerializeRunContext(root_task.flow):
+    with DematerializeRunContext(root_task.flow, allow_write_local_table_cache=False):
```
nicolasmueller, line 728: Should we also expose this as an option? I didn't do it for now, since I thought `get_output_from_store` is even more in debugging territory than `Result.get()`.

windiana42: I got feedback that getting inputs and outputs of tasks from currently cached state would be nice. Especially getting inputs might be combined with the ability to call a task outside of `with Flow` in case you can already provide the inputs dematerialized. Exposing write_local_table_cache there would make sense indeed. But we could also skip this for now and create a separate issue for this.

windiana42 (approved):
lgtm

src/pydiverse/pipedag/materialize/core.py
NicolasMuellerQC: I'll merge this PR so I'm unstuck but I can create another for `get_output_from_store`

PR #303, opened 2025-12-03
fix: Avoid recursive assumed dependencies

### LINKED ISSUE #304: Task with many imperatively materialized tables hangs
Currently
```python
def test_many_imperative_tables(lazy):
    @materialize(input_type=pd.DataFrame, lazy=lazy)
    def imperative_task():
        tables = []
        for i in range(50):
            df = pd.DataFrame({"counter": [i]})
            tables.append(Table(df).materialize())
        return Table(df)

    with Flow() as flow:
        with Stage("stage_1"):
            pd_tbl = imperative_task()
            m.noop(pd_tbl)
    flow.run()
```

hangs for both `lazy=True` and `False`. This is because we try to JSON serialize the `table.assumed_dependencies`, which results in a huge output, due to the recursive structure.

nicolasmueller:
Fixes #304.

# Checklist

### DIFF
--- a/src/pydiverse/pipedag/materialize/cache.py
+++ b/src/pydiverse/pipedag/materialize/cache.py
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import itertools
-from copy import copy
 from dataclasses import dataclass
 from functools import cached_property
 from typing import TYPE_CHECKING
@@ -27,7 +26,7 @@ class ImperativeMaterializationState:
         self.auto_suffix_counter = itertools.count()
 
     def add_table_lookup(self, obj, table: Table):
-        table_without_assumed_dependencies = copy(table)
+        table_without_assumed_dependencies = table.copy_without_obj()
         table_without_assumed_dependencies.assumed_dependencies = None
         self.assumed_dependencies.add(table_without_assumed_dependencies)
         self.object_lookup[id(obj)] = table


src/pydiverse/pipedag/materialize/cache.py
```diff
@@ -26,7 +27,9 @@ def __init__(self):
         self.auto_suffix_counter = itertools.count()
 
     def add_table_lookup(self, obj, table: Table):
-        self.assumed_dependencies.add(table)
+        table_without_assumed_dependencies = copy(table)
```
windiana42: My gut feeling would prefer table.copy_without_obj(). But I don't have a good argument why copy(table) is not just as good. It still keeps references to the same substructures. It must be clear that if table.cache_key is later modified it will not update here. But this should not matter since the imperative materialization should already have set the cache_key before the copy.
So my preference for a deep copy without obj probably comes from the fact that you don't need to *think* twice which references might be shared and later modified even though there is none.

docs/source/changelog.md
```diff
@@ -2,6 +2,7 @@
 
 ## 0.12.6 (2025-XX-XX)
 - Feat: Automatically check cache-validity of polars and pandas DataFrame tasks marked as lazy
+- Fix: Fix hang when using many imperative materializations in a single task
```
windiana42:
```suggestion
- Fix: Fix hang when using many imperative materializations in a single task (the implementation was not designed for this and storage size grew unsustainably)
```

PR #305, opened 2025-12-03
fix: Hash table names

nicolasmueller:
Hash table names as part of query hash. This avoids a crash when only renaming a lazy table that results from a dataframe.

# Checklist

### DIFF
--- a/tests/test_cache/test_basic_cache_invalidation.py
+++ b/tests/test_cache/test_basic_cache_invalidation.py
@@ -1114,9 +1114,13 @@ def test_broken_df_hashing(mocker):
         res_pd_spy.assert_called_once()
 
 
-def test_lazy_dataframe_table_name_change():
+@pytest.mark.parametrize("imperative", [False, True])
+def test_lazy_dataframe_table_name_change(imperative):
     @materialize(input_type=pd.DataFrame, lazy=True)
     def get_df():
+        if imperative:
+            for i in range(2):
+                Table(pd.DataFrame({"x": [i]}), name=f"{name}_{i}").materialize()
         return Table(pd.DataFrame({"x": [0]}), name=name)
 
     for i in range(2):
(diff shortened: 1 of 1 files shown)


windiana42:
This code is to blame:
```python
            ctx.defer_table_store_op(
                stage,
                DeferredTableStoreOp(
                    "_rename_table",
                    DeferredTableStoreOp.Condition.ON_STAGE_ABORT,
                    (from_name, table.name, from_schema),
                ),
            )
```
The idea is that if a stage is 100% cache valid. And if the only change is that a table name changed, then we simply rename the top level reference to this table and are good. Whenever we use stage_commit_technique read_views, we should actually call _rename_view instead.

see also:
```python
    if dialect in ["duckdb", "ibm_db_sa"]:
        stage_commit_technique = "read_views"
    else:
        stage_commit_technique = "schema_swap"
```

windiana42:
I probably will create a new PR with the change for implementing the view renaming correctly. It uses the test function from this PR though but extends it a bit to trigger failures when resulting tables cannot be read any more:
```
def test_lazy_dataframe_table_name_change():
    logger = structlog.get_logger(__name__ + ".test_lazy_dataframe_table_name_change")
    name = None  # will be set before calling tasks

    @materialize(lazy=True)
    def get_df():
        return Table(pd.DataFrame({"x": [0]}), name=name)

    @materialize(lazy=True)
    def get_tbl():
        return Table(sa.select(sa.literal(1).label("x")), name=f"sql_{name}")

    for i in range(2):
        logger.info("### Running flow iteration", iteration=i)
        name = f"tbl_{i}"
        with Flow("test") as flow:
            with Stage("first"):
                df = get_df()
                tbl = get_tbl()
            with Stage("second"):
                _ = m.noop(df), m.noop(tbl), m.noop_lazy(df), m.noop_lazy(tbl)
        result = flow.run()
        assert result.successful
        if i > 0:
            assert result.task_states[result.flow["first"]["get_df"]] == FinalTaskState.CACHE_VALID
            assert result.task_states[result.flow["first"]["get_tbl"]] == FinalTaskState.CACHE_VALID
```

PR #343, opened 2026-05-26
Print original error messages when using fallbacks

DominikZuercherQC:
I had a hard time debugging issues with `bcp` because the logs did not include the original error messages when falling back to pandas/polars here.
Could we add the original error messages to the logs?

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -1,5 +1,7 @@
 # Changelog
 
+- Feat: #343 When encountering error in table download outputs original error as part of the failure message now.
+
 ## 0.12.13 (2026-04-16)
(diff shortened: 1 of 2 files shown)


nicolasmueller (approved):
Can you add a line to the changelog?

docs/source/changelog.md
```diff
@@ -1,5 +1,7 @@
 # Changelog
 
+- Feat: #343 When encountering error in table download outputs original error as part of the failure message now.
```
nicolasmueller:
```suggestion
- Feat: #343 When encountering error in MSSQL table download outputs original error as part of the failure message now.
```

